PR #12, opened 2023-02-24
feat: hide header icon actions for small screen

hel-axelor:
RM#61019


packages/core/src/components/organisms/HeaderOptionsMenu/HeaderOptionsMenu.js
```diff
@@ -70,9 +87,11 @@ const HeaderOptionsMenu = ({
             modelId,
           })
         }
-      />
-      <HeaderOptionMenuItem
+      />,
+      <MenuItem
+        key={1}
```
lme-axelor, line 92: Both items have the same key, is it normal?

PR #13, opened 2023-02-27
feat: add event planning screen

gca-axelor:
RM#60702

### DIFF
--- a/packages/apps/crm/src/screens/event/EventPlanningScreen.js
+++ b/packages/apps/crm/src/screens/event/EventPlanningScreen.js
@@ -51,7 +51,12 @@ function EventPlanningScreen({navigation}) {
   const fetchEventFilter = useCallback(
     searchValue => {
       setFilter(searchValue === '' ? null : searchValue);
-      dispatch(fetchPlannedEvent({date: dateSave, searchValue: searchValue}));
+      dispatch(
+        fetchPlannedEvent({
+          date: dateSave,
+          searchValue: searchValue === '' ? null : searchValue,
+        }),
+      );
     },
     [dispatch, dateSave],
   );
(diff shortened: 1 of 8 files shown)


packages/apps/crm/src/screens/event/EventPlanningScreen.js
```diff
@@ -0,0 +1,259 @@
...
+    return (
+      <TouchableOpacity activeOpacity={0.9} onPress={() => navigateToEvent(id)}>
+        <Card
+          key={id}
+          style={[styles.container, rendBorderColor(event.border)]}>
+          <Text style={styles.bold}>{event.subject}</Text>
+        </Card>
+      </TouchableOpacity>
+    );
+  };
```
lme-axelor, line 128: Those two functions render the same card with two optional fields, I think we can simplify by creating a PlanningEventCard component with a boolean as props to know if location and contactPartner should be rendered or not.

packages/apps/crm/src/screens/event/EventPlanningScreen.js
```diff
@@ -0,0 +1,259 @@
...
+    return getStyles(Colors);
+  }, [Colors]);
+
+  const fetchItemsByMonth = useCallback(
+    date => {
+      dateSave === null && setDateSave(date);
+      dispatch(fetchPlannedEvent({date: date}));
+    },
+    [dispatch, dateSave],
+  );
```
lme-axelor, line 76: Client filter disappear on refresh. fetchEventFilter should set the searchValue in a filter variable and this value should be re-used here to keep filter.

lme-axelor (changes requested):

lme-axelor (approved):

PR #19, opened 2023-03-06
fix: fix broken endpoints for aos v7

hel-axelor:
RM#61360

### DIFF
--- a/packages/core/src/auth/api/config-api.js
+++ b/packages/core/src/auth/api/config-api.js
@@ -17,17 +17,18 @@
  */
 
 import {axiosApiProvider} from '../../apiProviders';
+import {RouterProvider} from '../../config';
 
 export async function getBaseConfig() {
   return axiosApiProvider.get({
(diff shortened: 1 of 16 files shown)


packages/core/src/auth/api/config-api.js
```diff
@@ -20,7 +20,7 @@ import {axiosApiProvider} from '../../apiProviders';
 
 export async function getBaseConfig() {
   return axiosApiProvider.get({
-    url: '/ws/rest/com.axelor.apps.base.db.AppBase/',
+    url: '/ws/rest/com.axelor.studio.db.AppBase/',
```
lme-axelor, line 23: You forgot to use the router here. I updated your branch, can you check that all API requests are taken care of please ? It seems that a few are still missing. I saw '/ws/rest/com.axelor.apps.base.db.App/search' and '/ws/rest/com.axelor.apps.base.db.AppMobileSettings/' but maybe there is more.

PR #36, opened 2023-03-28
feat: Save one session for user

gca-axelor:
RM#62284


example/app.config.js
```diff
@@ -33,4 +33,9 @@ export const app_config = {
    * can be rewritten for any request using props numberElementsByPage
    */
   defaultRequestLimit: 10,
+
+  /*
+   * This setting is used to define the default mode of session connection
+   */
```
lme-axelor, line 39: A bit more explanation could be nice 

"This configuration allows you to activate or not the connection sessions. If enabled, the user will be able to save sessions on their device for later use."

packages/core/src/components/organisms/index.js
```diff
@@ -29,3 +29,4 @@ export {default as PasswordInput} from './PasswordInput/PasswordInput';
 export {default as ScannerAutocompleteSearch} from './ScannerAutocompleteSearch/ScannerAutocompleteSearch';
 export {default as UrlInput} from './UrlInput/UrlInput';
 export {default as UsernameInput} from './UsernameInput/UsernameInput';
+export {default as SessionInput} from './SessionInput/SessionInput';
```
lme-axelor, line 32: alphabetical order

packages/core/src/screens/LoginScreen.js
```diff
@@ -118,6 +131,21 @@ const LoginScreen = ({route}) => {
...
+      enableConnectionSessions &&
+      sessionStorage.insert({
+        data: {
+          id: 1,
+          url: url,
+          username: username,
+          name: sessionName,
+        },
+      });
+    sessionStorage.recrypt({modelName: 'my-encryption-key!'});
```
lme-axelor, line 146: to be removed

packages/core/src/utils/session.ts
```diff
@@ -0,0 +1,23 @@
...
+    this.recrypt();
+  }
+
+  insert({data}: {data: any}) {
+    this.localStorage.setItem(this.key, data);
+  }
+  getItem() {
+    const item = this.localStorage.getItem(this.key);
+    return item;
+  }
```
lme-axelor, line 17: As this is an interface to get sessions, I think we could put function names with more meaning
addSession
getSession

Also, please add type for a session object to define each needed props, I think I added it in the ticket..

PR #48, opened 2023-04-17
feat: add translations and remove success code in toast message

gca-axelor:
RM#63110


packages/core/src/apiProviders/utils.ts
```diff
@@ -71,8 +68,10 @@ const manageError = (
         type: 'error',
         position: 'bottom',
         bottomOffset: 20,
-        text1: `Error ${code}`,
```
lme-axelor, line 74: We should keep code for error message

packages/core/src/apiProviders/utils.ts
```diff
@@ -71,8 +68,10 @@ const manageError = (
...
-        text2: `Failed to ${action}: ${message}.`,
+        text1: i18nProvider.i18n.t('Base_Error'),
+        text2: ` ${i18nProvider.i18n.t(
+          'Base_Failed_To',
+        )} ${action}: ${message}.`,
```
lme-axelor, line 74: As action is something given in function props, I think we can do better : 
- ask in props for a translation key
- display translated key

Caution: this implies changing all use of this function by transforming string to translation key

packages/core/src/apiProviders/utils.ts
```diff
@@ -104,8 +102,10 @@ const manageSucess = (
       type: 'success',
       position: 'bottom',
       bottomOffset: 20,
-      text1: `Success ${code}`,
-      text2: `${message ? message : 'Request successful'}.`,
+      text1: i18nProvider.i18n.t('Base_Success'),
+      text2: `${
+        message ? message : i18nProvider.i18n.t('Base_Request_Successful')
+      }.`,
```
lme-axelor, line 108: In case no message was given for request response, I think it could be better to have the same system as error toast by displaying the action instead of "Request successful" to give more informations on action.

"Successfully ${action}" / "${action} réussi(e)" -> This implies french translation should be something like "Mise à jour de la ligne" or "Ajout de la ligne"

PR #54, opened 2023-05-05
feat: enable to define list of headband to display above header

hel-axelor:
RM#62665


lme-axelor (commented):
Having a context to manage header bands is quite overkill I think, a provider would have been sufficient but It's working so why not ! 
I just think we should simplify the context : **only one band list** with all the header bands. _The component HeaderBandList should filter the bands itself before displaying._
Also, I think we should have **only one function** to manage bands: _registerHeaderBand_ which should be able to add a new band when key doesn't exist, update an existing band (all props should be editable). This way, to remove a band, we only have to register the same band with a condition to false.

hel-axelor:
I tried to use a provider like headerActionProvider. But, it doesn't track the header band state because the HeaderBandList component is at the top of the application hierarchic and the action changes are in depth components, So there is one solution is to use context.

PR #55, opened 2023-05-05
feat: add banner on header when user is connected on test instance

hel-axelor:
RM#62668


packages/core/src/app/ContextedApplication.tsx
```diff
@@ -109,7 +109,7 @@ const ContextedApplication = ({
       <Translator />
       <ErrorBoundary
         errorScreen={ErrorScreen}
-        userIdfetcher={getActiveUserId}
```
lme-axelor, line 112: Instead of modifying ErrorBoundary object, I think it could be a better practice to define a callback function which return the same format as before using getActiveUserInfo

`
const getActiveUserId = useCallback(() => getActiveUserInfo().then(({userId}) => userId), [...]);
`

PR #56, opened 2023-05-09
feat: save url in storage

gca-axelor:
RM#64140

### DIFF
--- a/packages/core/src/features/authSlice.js
+++ b/packages/core/src/features/authSlice.js
@@ -19,6 +19,7 @@
 import {createAsyncThunk, createSlice} from '@reduxjs/toolkit';
 import axios from 'axios';
 import {getActiveUserId, loginApi} from '../api/login-api';
+import {storage} from '../storage/Storage';
 import {testUrl} from '../utils/api';
 
 export const login = createAsyncThunk(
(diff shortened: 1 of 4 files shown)


packages/core/src/screens/LoginScreen.js
```diff
@@ -76,6 +78,9 @@ const LoginScreen = ({route}) => {
   }, [modeDebug, releaseInstanceConfig?.showUrlInput]);
 
   const defaultUrl = useMemo(() => {
+    if (urlStorage !== undefined) {
+      return urlStorage;
+    }
```
lme-axelor, line 83: It's better to check for undefined and null values by doing != null instead of !== undefined

```
    if (urlStorage != null) {
      return urlStorage;
    }
```

packages/core/src/features/authSlice.js
```diff
@@ -31,6 +32,7 @@ export const login = createAsyncThunk(
       password,
     );
     const userId = await getActiveUserId();
+    storage.setItem('url', urlWithProtocol);
```
lme-axelor, line 35: A better practice is to give meaning to keys, just 'url' does not really explain what it is used for. Also, it could be better to define this key only one time with an exported variable and use this variable in login screen:

```
   const URL_STORAGE_KEY = 'LastConnectionUrl';
...
    storage.setItem(URL_STORAGE_KEY, urlWithProtocol);
```

PR #59, opened 2023-05-16
feat:  manage drawer subtitles dynamically

gca-axelor:
RM#64352

### DIFF
--- a/packages/core/src/auth/screens/SettingsScreen.js
+++ b/packages/core/src/auth/screens/SettingsScreen.js
@@ -47,6 +47,8 @@ const SettingsScreen = ({route, children}) => {
     setActivityIndicator,
     toggleFilterConfig,
     toggleVirtualKeyboardConfig,
+    setShowSubtitles,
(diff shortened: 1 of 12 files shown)


packages/core/src/auth/screens/SettingsScreen.js
```diff
@@ -88,6 +90,10 @@ const SettingsScreen = ({route, children}) => {
     [Theme],
   );
 
+  const handleToggleSubtitles = useCallback(() => {
+    setShowSubtitles(!showSubtitles);
+  }, [setShowSubtitles, showSubtitles]);
```
lme-axelor, line 95: this is not optimized, each time the function is called, it will re-render the function as dependencies are modified.
You can get current value of showSubtitles directly is setShowSubtitles function : 

```
  const toggleSubtitles = useCallback(() => {
    setShowSubtitles(_currentState =>!_currentState);
  }, [setShowSubtitles]);
```

PR #66, opened 2023-05-30
chore: update CI 6.5

gca-axelor:


lme-axelor:
Used #65 with cherry-pick

PR #68, opened 2023-06-02
feat: add progressbar component

gca-axelor:
RM#64911

### DIFF
--- a/packages/ui/src/components/organisms/ProgressBar/ProgressBar.tsx
+++ b/packages/ui/src/components/organisms/ProgressBar/ProgressBar.tsx
@@ -56,7 +56,7 @@ const ProgressBar = ({
   if (showPercent) {
     displayValue = `${Math.round(percent)}%`;
   } else {
-    displayValue = `${value}/${total}`;
+    null;
   }
 
   let color: Color = Colors.cautionColor;
--- a/packages/ui/src/components/organisms/ProgressBar/ProgressBar.tsx
+++ b/packages/ui/src/components/organisms/ProgressBar/ProgressBar.tsx
@@ -1,0 +1,108 @@
+/*
+ * Axelor Business Solutions
+ *
+ * Copyright (C) 2023 Axelor (<http://axelor.com>).
+ *
+ * This program is free software: you can redistribute it and/or  modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React, {useMemo} from 'react';
+import {StyleSheet, View} from 'react-native';
+import {ThemeColors, useThemeColor, Color} from '../../../theme';
+import {Text} from '../../atoms';
+
+type ProgressBarProps = {
+  value?: number;
+  total?: number;
+  showPercent?: boolean;
+  colorRepartition?: {[key: number]: Color};
+  height?: number;
+  style?: object;
+  styleTxt?: object;
+};
+
+const ProgressBar = ({
+  value = 0,
+  total = 100,
+  showPercent = true,
+  colorRepartition = {},
+  height = 30,
+  style,
+  styleTxt,
+}: ProgressBarProps) => {
+  const Colors = useThemeColor();
+  const percent = total !== 0 ? (value / total) * 100 : 0;
+  if (Object.keys(colorRepartition).length === 0) {
+    colorRepartition = {
+      0: Colors.errorColor,
+      25: Colors.cautionColor,
+      50: Colors.progressColor,
+      75: Colors.priorityColor,
+      100: Colors.successColor,
+    };
+  }
+
+  let displayValue;
+  if (showPercent) {
+    displayValue = `${Math.round(percent)}%`;
+  } else {
+    displayValue = `${value}/${total}`;
+  }
+
+  let color: Color = Colors.cautionColor;
+  Object.keys(colorRepartition).forEach(key => {
+    const threshold = parseInt(key, 10);
+    if (percent >= threshold && colorRepartition[threshold]) {
+      color = colorRepartition[threshold];
+    }
+  });
+  const styles = useMemo(
+    () => getStyles(Colors, height, color, percent),
+    [Colors, height, color, percent],
+  );
+
+  return (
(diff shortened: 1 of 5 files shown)


gca-axelor:
infinite strip dont work on storybook but other props work very fine

packages/ui/src/components/organisms/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,173 @@
...
+      shadowColor: Colors.secondaryColor.background,
+      shadowOffset: {width: 0, height: 0},
+      backgroundColor: Colors.backgroundColor,
+    },
+    progressBar: {
+      borderRadius: 7,
+      justifyContent: 'center',
+      alignItems: 'center',
+      backgroundColor: color.background,
+      width: `${percent}%`,
```
lme-axelor, line 156: width should not be over 100% even if percent is over 100 otherwise text is getting out of the screen
Use min value between percent and 100 here

packages/ui/src/components/organisms/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,173 @@
...
+    container: {
+      width: '100%',
+      height: height,
+      borderRadius: 7,
+      elevation: 3,
+      shadowOpacity: 0.5,
+      shadowColor: Colors.secondaryColor.background,
+      shadowOffset: {width: 0, height: 0},
+      backgroundColor: Colors.backgroundColor,
+    },
```
lme-axelor, line 150: Please add a small border on container with secondaryColor

packages/ui/src/components/organisms/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,173 @@
...
+                transform: [{translateX}],
+              },
+            ]}
+          />
+        )}
+        {showPercent && (
+          <Text style={[styles.text, styleTxt]}>{`${Math.round(
+            percent,
+          )}%`}</Text>
+        )}
```
lme-axelor, line 127: Not as described in the ticket
props showPercent should be used to know if the value displayed should be {percent}% OR {value}/{total}
There should always have a value displayed.

lme-axelor, line 127: There is also an issue when percent is 0 as there is no text displayed and 5 as test is cut, text could be displayed in the container instead of in the progressBar when text does not fit

packages/ui/src/components/organisms/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+  if (Object.keys(colorRepartition).length === 0) {
+    colorRepartition = {
+      0: Colors.errorColor,
+      25: Colors.cautionColor,
+      50: Colors.progressColor,
+      75: Colors.priorityColor,
+      100: Colors.successColor,
+    };
+  }
```
lme-axelor, line 62: Please wrap this in an useMemo

```
  const progressColors = useMemo(() => {
           if (Object.keys(colorRepartition).length === 0) {
                 return {
                      0: Colors.errorColor,
                      25: Colors.cautionColor,
                      50: Colors.progressColor,
                      75: Colors.priorityColor,
                     100: Colors.successColor,
               }
          }
          return colorRepartition;
    }, [...]);
```

packages/ui/src/components/organisms/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,173 @@
...
+  colorRepartition = {},
+  stripe = true,
+  height = 30,
+  stripeDuration = 1000,
+  stripeWidth = 40,
+  style,
+  styleTxt,
+}: ProgressBarProps) => {
+  const Colors = useThemeColor();
+  const percent = total !== 0 ? (value / total) * 100 : 0;
```
lme-axelor, line 50: useMemo for this too please

packages/ui/src/components/organisms/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,173 @@
...
+    stripeAnimation.start();
+  }, [animatedStripe, stripeDuration]);
+
+  const translateX = animatedStripe.interpolate({
+    inputRange: [0, 1],
+    outputRange: [-stripeWidth, Dimensions.get('window').width],
+  });
+
+  return (
+    <View style={[styles.container, style]}>
```
lme-axelor, line 102: Use onLayout props to get view width and use it in stripe animation

packages/ui/src/components/organisms/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,173 @@
...
+  useEffect(() => {
+    const stripeAnimation = Animated.loop(
+      Animated.timing(animatedStripe, {
+        toValue: 1,
+        duration: stripeDuration,
+        useNativeDriver: true,
+      }),
+    );
+    stripeAnimation.start();
+  }, [animatedStripe, stripeDuration]);
```
lme-axelor, line 94: This animation is broken when user reset progress after it reached 100%

packages/ui/src/components/organisms/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+  const renderPercent = useCallback(() => {
+    if (showPercent) {
+      return (
+        <Text style={[styles.text, styleTxt]}>{`${percent.toFixed(2)}%`}</Text>
+      );
+    } else {
+      return <Text style={[styles.text, styleTxt]}>{`${value}/${total}`}</Text>;
+    }
+  }, [percent, showPercent, styles, styleTxt, total, value]);
```
lme-axelor, line 117: Still some issue on text display, when outside of progress bar it should be on the close to progress border as a mirror effect of the normal layout

And the text is still cut on some value, if think disposition should be based on text length AND progress width instead of percent value

packages/ui/src/components/organisms/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,196 @@
...
+        useNativeDriver: true,
+      }),
+    );
+    stripeAnimation.start();
+  }, [animatedStripe, stripeDuration, value]);
+
+  const translateX = animatedStripe.interpolate({
+    inputRange: [0, 1],
+    outputRange: [-stripeWidth, progressBarWidth],
+  });
```
lme-axelor, line 107: Still broken :grimacing:

packages/ui/src/components/organisms/ProgressBar/ProgressBar.tsx
lme-axelor: Maybe we can add a props to know if value should always be centered in container or if it should follow progress bar (idea from Marc :innocent: )

Also, progress bar is not centered anymore in container

PR #75, opened 2023-06-08
Wip feat 64919

gca-axelor:
RM#64919


Gregory256c:
NEED TO DEBUG FORM COMPONENT

gca-axelor:
KEYBOARD AVOIDING PB

packages/apps/helpdesk/src/screens/TicketFormScreen.js
```diff
@@ -0,0 +1,320 @@
...
+            style={styles.picker}
+            styleTxt={styles.marginTitle}
+          />
+          <FormIncrementInput
+            title={I18n.t('Helpdesk_Progress')}
+            defaultValue={_ticket.progressSelect?.toString()}
+            onChange={value => handleTicketFieldChange(value, 'progressSelect')}
+            decimalSpacer={I18n.t('Base_DecimalSpacer')}
+            thousandSpacer={I18n.t('Base_ThousandSpacer')}
+          />
```
lme-axelor, line 161: Progress is not an decimal value tike others. You can test on the web instance if you want, we can only set particular interger value so I think we should developp a specific component in Helpdesk module : ProgressFormInput which should have the same design as FormIncrementInput but formats the value entered to the nearest decimal place, and an attribute should be added to the Increment component to disable automatic formatting (formatting must be enabled by default).

packages/apps/helpdesk/src/features/ticketSlice.js
```diff
@@ -89,10 +91,46 @@ export const updateTicketStatus = createAsyncThunk(
...
+
+export const updateTicket = createAsyncThunk(
+  'ticket/updateTicket',
+  async function (data = {}, {getState}) {
+    return handlerApiCall({
+      fetchFunction: _updateTicket,
+      data,
+      action: 'Heldesk_Update_Ticket',
+      getState,
+      responseOptions: {isArrayResponse: false},
```
lme-axelor, line 115: show success toast here please

PR #85, opened 2023-06-15
fix: filter product stocks with user active stock location

hel-axelor:
- set stock location search bar default value for the product details screen.
-  fix auto complete exceede renders.

RM#65228


hel-axelor:
We should evaluate the stability of the **Scanner Auto Complete** when assigning it a default value, as it has been observed to show inconsistent behavior.

PR #86, opened 2023-06-16
fix: internal move create screen fields visibilty

hel-axelor:
RM#65130

### DIFF
--- a/packages/apps/stock/src/screens/internalMoves/InternalMoveLineCreationScreen.js
+++ b/packages/apps/stock/src/screens/internalMoves/InternalMoveLineCreationScreen.js
@@ -68,6 +68,12 @@ const InternalMoveLineCreationScreen = ({navigation}) => {
 
   const handleToProductTrackingNumberChange = useCallback(
     _value => {
+      if (destinationStockLocation != null) {
+        setStockLocationLine(_value);
+        setMovedQty(0);
+        return;
+      }
+
       if (_value == null) {
         handleReset(CREATION_STEP.product_trackingNumber);
       } else {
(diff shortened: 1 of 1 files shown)


packages/apps/stock/src/screens/internalMoves/InternalMoveLineCreationScreen.js
```diff
@@ -68,14 +68,19 @@ const InternalMoveLineCreationScreen = ({navigation}) => {
 
   const handleToProductTrackingNumberChange = useCallback(
     _value => {
+      if (destinationStockLocation != null) {
+        setStockLocationLine(_value);
+        return;
+      }
+
```
lme-axelor, line 75: I don't understand the functional meaning of this. I don't think this is a good pratice too as you are skipping checks

hel-axelor: This is to skip the next step when we change the product, ensuring that the selected destination stock location is not lost.

packages/apps/stock/src/screens/internalMoves/InternalMoveLineCreationScreen.js
```diff
@@ -68,14 +68,20 @@ const InternalMoveLineCreationScreen = ({navigation}) => {
 
   const handleToProductTrackingNumberChange = useCallback(
     _value => {
+      if (destinationStockLocation != null) {
+        setStockLocationLine(_value);
+        setMovedQty(0);
+        return;
+      }
+
```
lme-axelor, line 76: This has still no functional meaning ! **Do not skip checks !**
The changes needed here have to be done in helper functions (`handleReset` & `handleNextStep`).
If checks are there, it is for a reason, you cannot skip them whenever you want or you will lose functionality (`currentStep` value change for exemple here...). Moreover `handleReset` & `handleNextStep` enable to standardize the code and make it more readable : 
* If value is null then we are currently doing a reset of the step --> functional behaviour of reset of this step is done in `handleReset`, not at two places at the same time...
* If value is not null, then we need to go to next step --> functionnal behaviour of going to next step is done in `handleNextStep`, nowhere else

PR #88, opened 2023-06-27
feat: Manage config of number decimal digits

gca-axelor:
RM#65045


packages/core/src/hooks/use-digit-format.js
```diff
@@ -0,0 +1,34 @@
...
+  const formatDigit = useCallback(
+    numberToFormat => {
+      return parseFloat(numberToFormat.toString()).toFixed(
+        nbDecimalDigitForQty,
+      );
+    },
+    [nbDecimalDigitForQty],
+  );
+  return formatDigit;
+};
```
lme-axelor, line 34: Please move this to UI package

PR #89, opened 2023-06-27
fix:Check for NPE and null string condition in CRM

gca-axelor:
RM#66070


lme-axelor (commented):
Missing : DropdownGeneralView, EventHeader, EventLabelsCard, LeadHeader,

PR #106, opened 2023-07-26
refactor: simplify helpdesk object cards

gca-axelor:
RM#66785


gca-axelor:
ObjectCard : 6d28b0de296b38c215dcca163573c97ea163332f

PR #113, opened 2023-08-04
feat:  add contact provider to register phone contact

hel-axelor:
RM#66970


packages/core/src/tools/ContactProvider.ts
```diff
@@ -0,0 +1,133 @@
...
+  NEVER_ASK_AGAIN = 'never_ask_again',
+}
+
+class ContactProvider {
+  constructor() {}
+
+  private _requestReadContactsPermission =
+    async (): Promise<PermissionResult> => {
+      try {
+        const granted = await PermissionsAndroid.request(
```
lme-axelor, line 46: We will need this tool on iOS too, can you check if there are permissions to add please ? I will test them after on the iOS environment

packages/core/src/tools/ContactProvider.ts
```diff
@@ -0,0 +1,133 @@
...
+      }
+    };
+
+  private _createContactData = (
+    contactData: ContactData,
+  ): Partial<Contacts.Contact> => {
+    return {
+      emailAddresses: [
+        {
+          label: 'contact email address',
```
lme-axelor, line 68: If it is the display test associated with the email, maybe 'email' should sufficient ?

packages/core/src/tools/ContactProvider.ts
```diff
@@ -0,0 +1,133 @@
...
+          return 'Contacts permission denied.';
+        case PermissionResult.NEVER_ASK_AGAIN:
+          return 'Contacts permission denied permanently.';
+        default:
+          return 'Unknown permission result.';
+      }
+    } catch (error) {
+      console.error('Error adding contact:', error);
+      return 'Error adding contact: ' + error.message;
+    }
```
lme-axelor, line 129: Can you please show a toast message on cases : granted, error and denied. I also think that denied and never_ask_again could be in the same switch case here as the result is the same, contact was not created.

packages/core/src/tools/ContactProvider.ts
```diff
@@ -0,0 +1,133 @@
...
+    };
+  };
+
+  saveContact = async (contactData: ContactData): Promise<string> => {
+    try {
+      if (
+        !contactData.firstName ||
+        !contactData.lastName ||
+        !contactData.phoneNumbers.mobilePhone
+      ) {
```
lme-axelor, line 108: mobile phone number should not be required, only firstName should be required.

packages/core/src/tools/ContactProvider.ts
```diff
@@ -0,0 +1,133 @@
...
+      } catch (error) {
+        console.error('Error requesting contacts permission:', error);
+        return PermissionResult.DENIED;
+      }
+    };
+
+  private _createContactData = (
+    contactData: ContactData,
+  ): Partial<Contacts.Contact> => {
+    return {
```
lme-axelor, line 65: I think we should check presence of data in each field before adding them to the body as none of the data is required other than firstName. I'm afraid the created contact will be dirty if null values with label are given to contact addition tool. Please also modify ContactData interface to inform of optional values.

packages/core/src/tools/ContactProvider.ts
```diff
@@ -0,0 +1,133 @@
...
+}
+
+class ContactProvider {
+  constructor() {}
+
+  private _requestReadContactsPermission =
+    async (): Promise<PermissionResult> => {
+      try {
+        const granted = await PermissionsAndroid.request(
+          PermissionsAndroid.PERMISSIONS.READ_CONTACTS,
```
lme-axelor, line 47: Is it really required to ask for READ_CONTACTS permission when trying to add a contact ?

packages/core/src/tools/ContactProvider.ts
```diff
@@ -0,0 +1,133 @@
...
+          state: null,
+          postCode: null,
+          country: null,
+        },
+      ],
+      note: contactData.notes,
+    };
+  };
+
+  saveContact = async (contactData: ContactData): Promise<string> => {
```
lme-axelor, line 102: Instead of `Promise<string>` it could be more useful to return `Promise<boolean>` to know if addition was successful are not.

package.json
```diff
@@ -41,6 +41,7 @@
     "react": "18.2.0",
     "react-i18next": "^11.16.7",
     "react-native": "0.68.1",
+    "react-native-contacts": "^7.0.6",
```
lme-axelor, line 44: Normally you should not need this dependency here as the package has been added to "dependencies" of core package, only "peerDependencies" should be given to both core package and global package. 
If it's not working when removing this dependency here, please move "react-native-contacts" from dependencies to peerDependencies in core package.

packages/core/src/tools/ContactProvider.ts
hel-axelor: Yep, it was mentioned in the doc of the react-native-contacts.

packages/core/src/tools/ContactProvider.ts
```diff
@@ -0,0 +1,171 @@
...
+import {PERMISSIONS, Permission, request} from 'react-native-permissions';
+import * as Contacts from 'react-native-contacts';
+
+interface ContactData {
+  firstName: string;
+  lastName?: string;
+  phoneNumbers?: {
+    mobilePhone: string;
+    fixedPhone: string;
+  };
```
lme-axelor, line 30: Both phone numbers are not required. I think it should be better to just ask for mobilePhone and fixedPhone directly instead of wrapping them by phoneNumbers

```
interface ContactData {
  firstName: string;
  lastName?: string;
  mobilePhone?: string;
  fixedPhone?: string;
  email?: string;
  address?: string;
  notes?: string;
```

packages/core/src/tools/ContactProvider.ts
```diff
@@ -0,0 +1,171 @@
...
+              city: null,
+              region: null,
+              state: null,
+              postCode: null,
+              country: null,
+            },
+          ]
+        : undefined,
+      note: notes || undefined,
+    };
```
lme-axelor, line 126: I still find this construction quite dirty, I'm sure we can do something cleaner for that with a parser/mapper. It will be more maintainable and more flexible if we want to add a new field later. 

The following is working fine but I think we can do something even better by defining type of ContactEntry (phoneNumber, email, address, info) to replace the isArrayField/getData. What do you think ?
```
const CONTACT_MAPPER = {
  firstName: {
    _key: 'givenName',
    getData: (value: string) => value,
    isArrayField: false,
  },
  lastName: {
    _key: 'familyName',
    getData: (value: string) => value,
    isArrayField: false,
  },
  notes: {
    _key: 'note',
    getData: (value: string) => value,
    isArrayField: false,
  },
  mobilePhone: {
    _key: 'phoneNumbers',
    getData: (value: string) => ({
      label: 'mobile',
      number: value,
    }),
    isArrayField: true,
  },
  fixedPhone: {
    _key: 'phoneNumbers',
    getData: (value: string) => ({
      label: 'fixed',
      number: value,
    }),
    isArrayField: true,
  },
  email: {
    _key: 'emailAddresses',
    getData: (value: string) => ({
      label: 'email',
      email: value,
    }),
    isArrayField: true,
  },
  address: {
    _key: 'postalAddresses',
    getData: (value: string) => ({
      label: 'main address',
      formattedAddress: value,
      street: null,
      pobox: null,
      neighborhood: null,
      city: null,
      region: null,
      state: null,
      postCode: null,
      country: null,
    }),
    isArrayField: true,
  },
};

const parseContactData = (_contact: ContactData): Partial<Contacts.Contact> => {
  if (isEmpty(_contact)) {
    return {};
  }

  const result: Partial<Contacts.Contact> = {};

  Object.entries(CONTACT_MAPPER).forEach(([key, config]) => {
    if (_contact[key] != null) {
      const data = config.getData(_contact[key]);

      if (!config.isArrayField) {
        result[config._key] = data;
      } else {
        if (Array.isArray(result[config._key])) {
          result[config._key].add(data);
        } else {
          result[config._key] = [data];
        }
      }
    }
  });

  return result;
};
```

packages/core/src/tools/ContactProvider.ts
```diff
@@ -0,0 +1,171 @@
...
+      let toastMessage: string;
+
+      switch (permissionResult) {
+        case PermissionResult.GRANTED:
+        case PermissionResult.LIMITED:
+          const _contactData = this._createContactData(contactData);
+          await Contacts.addContact(_contactData);
+          toastType = 'success';
+          toastMessage = `Contact ${contactData.firstName} added successfully.`;
+          break;
```
lme-axelor, line 149: Even when granting access to contacts when asked, I got an error while creating contact : "java.lang.SecurityException: Permission denied writing com.android.provider.contacts.Contact..."

ios/AosMobile/Info.plist
```diff
@@ -1,91 +1,93 @@
...
+			<string>armv7</string>
+		</array>
+		<key>UISupportedInterfaceOrientations</key>
+		<array>
+			<string>UIInterfaceOrientationPortrait</string>
+		</array>
+		<key>UIViewControllerBasedStatusBarAppearance</key>
+		<false />
+		<key>NSContactsUsageDescription</key>
+		<string>This app would like to read your contacts.</string>
```
lme-axelor, line 91: iOS need a real descirption of why you need to access contact. The need is not to read contacts but to create contacts.

PR #128, opened 2023-08-25
feat: add header action to save contact for crm

hel-axelor:
RM#66971


packages/apps/crm/src/i18n/en.json
```diff
@@ -118,5 +118,6 @@
   "Crm_SliceAction_CreateOpportunity": "create opportunity",
   "Crm_SliceAction_FetchCompanyById": "get company by id",
   "Crm_SliceAction_CreateCatalog": "Create catalog",
-  "Crm_SliceAction_FetchProspectStatus": "fetch prospect status"
+  "Crm_SliceAction_FetchProspectStatus": "fetch prospect status",
+  "Crm_SaveContact": "Save contact"
```
lme-axelor, line 122: We try to keep translation files readable even with all the slice translations, so please follow the existing structure : 
first part of file should contain the normal translation that are displayed to user and the end of the file should contain the slice translations.

You translation should be above the SliceAction. 
Please modify fr and en files.

packages/apps/crm/src/hooks/use-crm-header-actions.ts
```diff
@@ -53,29 +54,81 @@ const useCatalogListActions = () => {
...
+              mobilePhone: client.mobilePhone,
+              fixedPhone: client.fixedPhone,
+              email: client.emailAddress?.address,
+              address: {
+                street:
+                  client.mainAddress?.addressL4 ?? client.mainAddress?.fullName,
+                country: client.mainAddress?.addressL7Country,
+                city: client.mainAddress?.primaryCity,
+                postCode: client.mainAddress?.zip,
+              },
```
lme-axelor, line 83: Those fields are not even fetched... It's a good idea to add them when saving contact but please when using a new field, add it to the objectFields file to fetch the informations. Otherwise it's useless to use it as information will always be null.

Same for all actions.

packages/apps/crm/src/hooks/use-crm-header-actions.ts
```diff
@@ -53,29 +54,81 @@ const useCatalogListActions = () => {
...
+              notes: client.notes,
```
lme-axelor, line 84: Client does not have notes, that's description, please check the props you are using.

packages/apps/crm/src/hooks/use-crm-header-actions.ts
```diff
@@ -153,15 +232,41 @@ const useOpportunityDetailsActions = () => {
...
 
   useEffect(() => {
     headerActionsProvider.registerModel('crm_prospect_details', {
       model: 'com.axelor.apps.base.db.Partner',
       modelId: prospect?.id,
       disableMailMessages: !mobileSettings?.isTrackerMessageEnabled,
       attachedFileScreenTitle: prospect?.simpleFullName,
+      actions: [
+        {
+          key: 'contact-saveContact',
```
lme-axelor, line 245:
```
          key: 'prospect-saveContact',
```

PR #171, opened 2023-10-06
feat: add icon to Button and modify style

vhu-axelor:
RM#69672


packages/ui/src/components/molecules/Button/Button.tsx
```diff
@@ -20,63 +20,109 @@ import React, {useMemo} from 'react';
...
   disabled?: boolean;
+  FontAwesome5?: boolean;
+  iconName?: string;
+  isNeutralBackground?: boolean;
+  onDisabledPress?: () => void;
+  onPress?: () => void;
+  style?: any;
+  styleTxt?: any;
+  title?: string;
+  width?: string;
```
lme-axelor, line 37: width should also accept a number
It can be a fixed width defined as a number or defined using `Dimensions.get('window').width` which is also a number

packages/ui/stories/molecules/Button.stories.tsx
```diff
@@ -33,31 +33,63 @@ storiesOf('ui/molecules/Button', module).add(
   },
   {
     argTypes: {
-      title: {
-        type: 'string',
-        defaultValue: 'Press me',
-        control: {type: 'text'},
-      },
       color: {
         options: Object.keys(lightTheme.colors),
```
lme-axelor, line 37: We should maybe remove colors which are directly a text and not a Color object. This has been done on InfoBubble story I think

packages/ui/stories/molecules/Button.stories.tsx
```diff
@@ -33,31 +33,63 @@ storiesOf('ui/molecules/Button', module).add(
...
+          type: 'object',
+        },
+        defaultValue: {},
+      },
+      styleTxt: {
+        control: {
+          type: 'object',
+        },
+        defaultValue: {},
+      },
```
lme-axelor, line 82: You do not need to add style in storybook, it's do complicated to fill in when you are on the story, the editor is not practical

packages/ui/stories/molecules/Button.stories.tsx
```diff
@@ -33,31 +33,63 @@ storiesOf('ui/molecules/Button', module).add(
...
+      title: {
+        type: 'string',
+        defaultValue: 'Press me',
+        control: {type: 'text'},
+      },
+      width: {
+        type: 'string',
+        defaultValue: '',
+        control: {type: 'text'},
+      },
```
lme-axelor, line 92: Please give default value here

PR #181, opened 2023-10-10
feat: creation of a new pop-up

vhu-axelor:
RM#69666


packages/ui/src/components/molecules/PopUp/NewPopUp.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+    return {
+      ...cancelButtonConfig,
+      title: cancelButtonConfig.title ?? translator('Base_Cancel'),
+      color: cancelButtonConfig.color ?? Colors.errorColor,
+      iconName: cancelButtonConfig.iconName ?? 'times',
+      hide: cancelButtonConfig.hide ?? false,
+      showInHeader: cancelButtonConfig.showInHeader ?? false,
+      width: cancelButtonConfig.width ?? 115,
+    };
```
lme-axelor, line 66: Simplify here by defining default config which will be rewritten by config given in props

```
    return {
      title: translator('Base_Cancel'), // Please give default value to translator or it might throw an error here : value => value
      color: Colors.errorColor,
      iconName: 'times',
      hide: false,
      showInHeader: false,
      width: 115, // No magic number please
      ...cancelButtonConfig,
    };
```

packages/ui/src/components/molecules/PopUp/NewPopUp.tsx
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    return {
+      ...confirmButtonConfig,
+      title: confirmButtonConfig.title ?? translator('Base_Ok'),
+      color: confirmButtonConfig.color ?? Colors.primaryColor,
+      iconName: confirmButtonConfig.iconName ?? 'check',
+      hide: confirmButtonConfig.hide ?? false,
+      width: confirmButtonConfig.width ?? 115,
+    };
```
lme-axelor, line 81: same here

packages/ui/src/components/molecules/PopUp/NewPopUp.tsx
```diff
@@ -0,0 +1,161 @@
...
+      visible={visible}
+      transparent
+      animationType="fade"
+      onRequestClose={_cancelButtonConfig?.onPress}>
+      <View style={styles.modalBackground}>
+        <Card style={[styles.container, style]}>
+          <View style={styles.headerContainer}>
+            {!!title && (
+              <Text writingType="title" fontSize={20}>
+                {translator ? translator(title) : title}
```
lme-axelor, line 95: I'm not sure translator is needed here, title will already be translated
If you want a translation key instead of a title, please rename in titleKey but I'm not sure this is needed here

packages/ui/src/components/molecules/PopUp/NewPopUp.tsx
```diff
@@ -0,0 +1,161 @@
...
+          {children}
+          <View style={styles.buttonsContainer}>
+            {!_cancelButtonConfig?.hide &&
+              !_cancelButtonConfig?.showInHeader && (
+                <Button {..._cancelButtonConfig} style={styles.cancelButton} />
+              )}
+            {!_confirmButtonConfig?.hide && (
+              <Button {..._confirmButtonConfig} />
+            )}
+          </View>
```
lme-axelor, line 118: If both buttons are hidden, this view should not be rendered as it will leave unwanted space

packages/ui/stories/molecules/NewPopUp.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+    argTypes: {
+      visible: {
+        type: 'boolean',
+        defaultValue: true,
+        control: {type: 'boolean'},
+      },
+      title: {
+        type: 'string',
+        defaultValue: 'Title',
+        control: {type: 'text'},
```
lme-axelor, line 58: It could be nice to add more props here to config buttons OR please add multiple stories to match each popup use cases (action in header, buttons hidden, only icons,...)

I think it's better to add configurable props

PR #189, opened 2023-10-16
fix: HR prevent user from completing informations depending on config

gca-axelor:
RM#68663

### DIFF
--- a/packages/apps/hr/src/models/forms.ts
+++ b/packages/apps/hr/src/models/forms.ts
@@ -164,6 +164,8 @@ export const hr_formsRegister: FormConfigs = {
         widget: 'increment',
         hideIf: ({objectState}) =>
           objectState.manageMode === ExpenseLine.modes.kilometric,
+        readonlyIf: ({objectState}) =>
+          objectState?.expenseProduct?.blockExpenseTax === true,
       },
       comments: {
         titleKey: 'Hr_Comments',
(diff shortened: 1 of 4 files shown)


gca-axelor:
functional test : 
Billets de train/metro/tramway => Unavailable to users : true
Billets d'avion => Block expense tax modification : true

packages/apps/hr/src/models/objectFields.ts
```diff
@@ -66,6 +66,8 @@ export const hr_modelAPI: ObjectFields = {
   hr_expenseType: schemaContructor.object({
     name: schemaContructor.string(),
     fullName: schemaContructor.string(),
+    blockExpenseTax: schemaContructor.boolean(),
+    unavailableToUsers: schemaContructor.boolean(),
```
vhu-axelor, line 70: I think it is not necessary to add this field

packages/apps/hr/src/models/forms.ts
```diff
@@ -164,6 +164,8 @@ export const hr_formsRegister: FormConfigs = {
         widget: 'increment',
         hideIf: ({objectState}) =>
           objectState.manageMode === ExpenseLine.modes.kilometric,
+        readonlyIf: ({objectState}) =>
```
vhu-axelor, line 167: Works on creation but not on update

PR #190, opened 2023-10-16
feat: add two new components for anomalies check

vhu-axelor:
RM#69869


packages/ui/src/components/molecules/InfoBubble/InfoBubble.tsx
```diff
@@ -77,7 +77,7 @@ const InfoBubble = ({
           size={coloredBubble ? size * 0.5 : size * 0.8}
         />
       </TouchableOpacity>
-      {isOpen ? (
+      {indication && isOpen ? (
```
lme-axelor, line 80: indication is not a boolean here, please make check more understandable

packages/core/src/utils/filter.ts
```diff
@@ -18,6 +18,29 @@
...
 import {Color} from '@axelor/aos-mobile-ui';
 
+interface Anomaly {
+  version: number;
+  message: string;
+  checkType: string;
+  modelName: string;
+}
+
+export function filterAnomaly(anomalyList: Anomaly[]) {
```
lme-axelor, line 28: That's not really a filter but more a sort function here

packages/core/src/utils/filter.ts
```diff
@@ -18,6 +18,29 @@
...
+  const BEFORE = -1;
+  const EQUAL = 0;
+  const AFTER = 1;
+
+  return anomalyList.sort((a: Anomaly, b: Anomaly) => {
+    if (a.checkType === b.checkType) {
+      return EQUAL;
+    } else if (a.checkType === 'error' && b.checkType === 'alert') {
```
lme-axelor, line 36: Those types 'error' & 'alert' are used more than one time, please create helper class in `types` folder to contain type values, color depending on type, sort function etc

packages/core/src/components/organisms/AnomalyList/AnomalyList.tsx
```diff
@@ -0,0 +1,81 @@
...
+  style?: any;
+}
+
+const AnomalyList = ({objectName, objectId, style}: AnomalyListProps) => {
+  const Colors = useThemeColor();
+
+  const [anomalyList, setAnomalyList] = useState([]);
+
+  useEffect(() => {
+    fetchAnomalies({objectName, objectId}).then(response => {
```
lme-axelor, line 37: there is not catch for error here, it should be to avoid problems

packages/core/src/components/organisms/AnomalyList/AnomalyList.tsx
```diff
@@ -0,0 +1,81 @@
...
+        const checks = response?.data?.object?.checks;
+        setAnomalyList(checks);
+      }
+    });
+  }, [objectName, objectId]);
+
+  const filteredAnomalyList = useMemo(
+    () => filterAnomaly(anomalyList),
+    [anomalyList],
+  );
```
lme-axelor, line 48: Instead of doing this, why not directly filtering list obtained for the API call before using the setState ?

packages/core/src/components/organisms/AnomalyBubble/AnomalyBubble.tsx
```diff
@@ -0,0 +1,90 @@
...
+        );
+        setAnomalyList([...checks, ...otherChecks]);
+      }
+    });
+  }, [objectName, objectId]);
+
+  const filteredAnomalyList = useMemo(
+    () => filterAnomaly(anomalyList),
+    [anomalyList],
+  );
```
lme-axelor, line 63: Same here

packages/core/src/components/organisms/AnomalyBubble/AnomalyBubble.tsx
```diff
@@ -0,0 +1,90 @@
...
+    return null;
+  }
+
+  return (
+    <View style={style}>
+      <InfoBubble
+        style={indicatorStyle}
+        iconName="exclamation-triangle"
+        badgeColor={
+          filteredAnomalyList[0].checkType === 'error'
```
lme-axelor, line 75: You are getting two times the same item from the list for display, I find it cleaner to use a useMemo here

PR #203, opened 2023-10-18
fix: export content of hr package to allow external package to use it

vhu-axelor:
RM#70558


packages/apps/hr/src/hooks/index.ts
```diff
@@ -0,0 +1,19 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+export {useHrHeaders} from './use-hr-header-actions';
```
lme-axelor, line 19: HR headers are already exported in module, that's not useful outside of core package and might cause issues if there are called at two different places. I think it's better to keep them inside the package only

PR #207, opened 2023-10-19
refactor: simplification of ChipSelect to use only one state for both…

vhu-axelor:
… modes
RM#70571


gca-axelor:
very sorry but not working functionally ❌❌

gca-axelor (approved):

PR #212, opened 2023-10-23
refactor: deletion of disabled props from Icon component

vhu-axelor:
RM#70400

### DIFF
--- a/packages/core/src/components/external/Stopwatch/Stopwatch.tsx
+++ b/packages/core/src/components/external/Stopwatch/Stopwatch.tsx
@@ -132,6 +132,31 @@ const Stopwatch = ({
     onCancel();
   };
 
+  const TimerButton = ({
+    name,
+    touchable,
+    onPress,
+  }: {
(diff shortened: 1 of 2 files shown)


gca-axelor (approved):

packages/core/src/components/external/Stopwatch/Stopwatch.tsx
```diff
@@ -143,55 +168,49 @@ const Stopwatch = ({
...
-                disablePlay ||
-                stopwatchStatus === StopwatchType.status.InProgress ||
-                stopwatchStatus === StopwatchType.status.Finished
+              touchable={
+                !(
+                  disable ||
+                  disablePlay ||
+                  stopwatchStatus === StopwatchType.status.InProgress ||
+                  stopwatchStatus === StopwatchType.status.Finished
+                )
```
lme-axelor, line 179: A props disabled instead of touchable in TimerButton would simplify those condition

PR #214, opened 2023-10-25
feat: toast open pop-up with the whole message

vhu-axelor:


gca-axelor (approved):

packages/core/src/components/external/Toast/Toast.tsx
```diff
@@ -0,0 +1,131 @@
...
+  StyleSheet.create({
+    error: {
+      borderLeftColor: Colors.errorColor.background,
+    },
+    errorAlert: {
+      borderLeftWidth: 7,
+      borderLeftColor: Colors.errorColor.background,
+      paddingLeft: 13,
+      paddingRight: 20,
+    },
```
lme-axelor, line 102: errorAlert & successAlert have the most of the same style, it could be nice to merge both in only one style

packages/core/src/components/external/Toast/ToastProvider.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  public setMessage(message: string) {
+    this.message = message;
+  }
+
+  public setType(type: string) {
+    this.type = type;
+  }
+}
```
lme-axelor, line 53: I think you can simplify by creating a object which contains the three props. This way you only have to call one setter/getter to set/get the whole object

gca-axelor:
Capture d’écran de 2023-11-02 13-06-34
no centered when samll text => tested in Hr / expense line when i delete an expense line

packages/core/src/components/external/Toast/Toast.tsx
```diff
@@ -0,0 +1,151 @@
...
+      <ErrorToast
+        {...props}
+        style={styles.toast}
+        contentContainerStyle={styles.toastContent}
+        text1Style={styles.title}
+        text2Style={styles.detail}
+        text2NumberOfLines={3}
+      />
+    ),
+  };
```
lme-axelor, line 78: I think there is a refresh issue on border color. When defining toast config, border color should not depends on useMemo. It should be clearly written : success = success color, eror = error color
At the moment, toast appear as the last toast color is border color is updated when you click on the toast

PR #236, opened 2023-11-07
feat: add test for RightIconButton

gca-axelor:
RM#71559

### DIFF
--- a/packages/ui/__tests__/components/molecules/RightIconButton.test.js
+++ b/packages/ui/__tests__/components/molecules/RightIconButton.test.js
@@ -36,12 +36,7 @@ describe('RightIconButton Component', () => {
   it('displays the title when provided', () => {
     const wrapper = shallow(<RightIconButton {...props} />);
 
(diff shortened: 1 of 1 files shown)


packages/ui/__tests__/components/molecules/RightIconButton.test.js
```diff
@@ -0,0 +1,54 @@
...
+    const wrapper = shallow(<RightIconButton {...props} />);
+
+    expect(wrapper.contains(props.icon)).toBe(true);
+  });
+
+  it('calls onPress when the button is pressed', () => {
+    const wrapper = shallow(<RightIconButton {...props} />);
+    const touchable = wrapper.find(TouchableOpacity);
+    touchable.simulate('press');
+    expect(props.onPress).toHaveBeenCalled();
```
lme-axelor, line 52: This test will always return true even if TouchableOpacity is disabled (because of ReactNative) so when you don't need to check args provided, please only check that TouchableOpacity is not disabled.

vhu-axelor (approved):

PR #240, opened 2023-11-07
feat: add SwitchCard test

gca-axelor:
RM#71605

### DIFF
--- a/packages/ui/__tests__/components/molecules/SwitchCard.test.js
+++ b/packages/ui/__tests__/components/molecules/SwitchCard.test.js
@@ -18,7 +18,7 @@
 
 import React from 'react';
 import {shallow} from 'enzyme';
-import {SwitchCard, Text, Switch} from '@axelor/aos-mobile-ui';
+import {SwitchCard, Text} from '@axelor/aos-mobile-ui';
 
 describe('SwitchCard', () => {
   it('should render without crashing', () => {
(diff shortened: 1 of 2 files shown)


packages/ui/__tests__/components/molecules/SwitchCard.test.js
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React from 'react';
+import {shallow} from 'enzyme';
+import {SwitchCard, Text} from '@axelor/aos-mobile-ui';
+
+describe('SwitchCard', () => {
```
vhu-axelor, line 23: Add props at the beginning to avoid set the same props in each test

packages/ui/__tests__/components/molecules/SwitchCard.test.js
```diff
@@ -0,0 +1,62 @@
...
+        title="Test"
+        defaultValue={defaultValue}
+        onToggle={() => {}}
+      />,
+    );
+
+    expect(wrapper.find('Switch').prop('isEnabled')).toBe(defaultValue);
+  });
+
+  it('calls onToggle when the switch is toggled', () => {
```
vhu-axelor, line 54: Here you don't test if onToggle is called when the switch is toggle, you verify that the right function is given to the component

vhu-axelor (approved):

PR #245, opened 2023-11-09
fix: manage AOP errors on API request

gca-axelor:
RM#71686


gca-axelor:
test : 
user: client
pwd : @axclient

vhu-axelor (approved):
Nice

PR #257, opened 2023-11-15
feat: add testInstance.config file

vhu-axelor:
RM#71957


vhu-axelor:
Use `git update-index --skip-worktree ./src/testInstance.config.js` to not track the new file

PR #260, opened 2023-11-16
feat: add test cases for HeaderContainer component

vhu-axelor:
RM#72024

### DIFF
--- a/packages/ui/__tests__/components/organisms/HeaderContainer.test.js
+++ b/packages/ui/__tests__/components/organisms/HeaderContainer.test.js
@@ -1,0 +1,78 @@
+/*
+ * Axelor Business Solutions
+ *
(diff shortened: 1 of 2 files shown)


packages/ui/__tests__/components/organisms/HeaderContainer.test.js
```diff
@@ -0,0 +1,78 @@
...
+    );
+
+    expect(wrapper.find('[testID="children"]').exists()).toBe(false);
+  });
+
+  it('should not render children if forceHideByDefault is true', () => {
+    const wrapper = shallow(<HeaderContainer {...props} forceHideByDefault />);
+
+    expect(wrapper.find('[testID="children"]').exists()).toBe(false);
+  });
```
lme-axelor, line 58: Need to add a test to check config management too with a spyOn on useConfig. It has already been done in another component

PR #267, opened 2023-11-21
feat: add label for groundForRefusal

gca-axelor:
RM#72154

### DIFF
--- a/packages/apps/hr/src/components/templates/ExpenseHeader/ExpenseHeader.tsx
+++ b/packages/apps/hr/src/components/templates/ExpenseHeader/ExpenseHeader.tsx
@@ -18,7 +18,13 @@
 
 import React from 'react';
 import {View, StyleSheet} from 'react-native';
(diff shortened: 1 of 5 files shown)


vhu-axelor (approved):

packages/apps/hr/src/components/templates/ExpenseHeader/ExpenseHeader.tsx
```diff
@@ -47,6 +53,9 @@ const ExpenseHeader = ({}) => {
           ? user?.activeCompany?.currency?.symbol
           : user?.activeCompany?.currency?.code
       }`}</Text>
+      {!checkNullString(expense?.groundForRefusal) && (
+        <Label message={expense?.groundForRefusal} type="error" />
```
lme-axelor, line 57: Maybe we can add a title before the refusal message "Ground for refusal: " / "Motif de refus: " maybe 🤷🏼

lme-axelor (commented):
Label is not taking the same width as ToggleSwitch, can you make all components of this header have the same width on this screen please

PR #269, opened 2023-11-21
feat: add city, and distance on kilometric expense line card

gca-axelor:
RM#72150

### DIFF
--- a/packages/apps/hr/src/i18n/en.json
+++ b/packages/apps/hr/src/i18n/en.json
@@ -44,7 +44,6 @@
   "Hr_ReasonRefusal": "Reason for refusal",
   "Hr_CityNotFound": "At least one of the two cities could not be located.",
   "Hr_KilometerUnit": "km",
-
   "Hr_SliceAction_FetchExpenseLines": "fetch expense lines",
   "Hr_SliceAction_SearchDraftExpense": "fetch draft expenses",
   "Hr_SliceAction_FetchExpense": "fetch expenses",
--- a/packages/apps/hr/src/i18n/en.json
+++ b/packages/apps/hr/src/i18n/en.json
@@ -43,6 +43,8 @@
   "Hr_Refuse": "Refuse",
   "Hr_ReasonRefusal": "Reason for refusal",
   "Hr_CityNotFound": "At least one of the two cities could not be located.",
+  "Hr_KilometerUnit": "km",
+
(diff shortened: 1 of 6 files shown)


vhu-axelor (changes requested):

packages/apps/hr/src/i18n/en.json
```diff
@@ -42,6 +42,8 @@
   "Hr_Refuse": "Refuse",
   "Hr_ReasonRefusal": "Reason for refusal",
   "Hr_CityNotFound": "At least one of the two cities could not be located.",
+  "Hr_KilometerUnit": "km",
+
```
lme-axelor, line 46: remove space, not needed in translations

packages/apps/hr/src/i18n/fr.json
```diff
@@ -42,6 +42,8 @@
   "Hr_Refuse": "Refuser",
   "Hr_ReasonRefusal": "Motif de refus",
   "Hr_CityNotFound": "Au moins une des deux villes n'a pas pu être localisée.",
+  "Hr_KilometerUnit": "km",
+
```
lme-axelor, line 46: remove space, not needed in translations

packages/ui/src/components/templates/ObjectCard/ObjectCard.tsx
```diff
@@ -165,6 +159,12 @@ const ObjectCard = ({
       return null;
     }
 
+    if (item.customComponent != null) {
+      return React.cloneElement(item.customComponent, {
+        key: `${item.displayText} - ${item.order}`,
+      });
+    }
+
```
lme-axelor, line 167: I'm not sure it's a good idea to move it this way. 
I understand the need of having the hideIf being evaluate as a first thing but the hideIfNull can cause issue with the custom component as displayText and indicatorText are not used. Maybe we can split condition into two parts like this : 

```
  if (item.hideIf) {
      return null;
    }

    if (item.customComponent != null) {
      return React.cloneElement(item.customComponent, {
        key: `${item.displayText} - ${item.order}`,
      });
    }

  if (item.hideIfNull &&
        checkNullString(item.displayText) &&
        checkNullString(item.indicatorText)) {
      return null;
    }
```

vhu-axelor (approved):

lme-axelor:
Need to be splited into two commits before merging to go down to version 7.1

PR #273, opened 2023-11-21
feat: add condition to display toogle depends on manage employee

gca-axelor:
RM#72153

### DIFF
--- a/packages/apps/hr/src/features/employeeSlice.js
+++ b/packages/apps/hr/src/features/employeeSlice.js
@@ -17,10 +17,7 @@
  */
 
 import {createAsyncThunk, createSlice} from '@reduxjs/toolkit';
-import {
-  generateInifiniteScrollCases,
-  handlerApiCall,
-} from '@axelor/aos-mobile-core';
+import {handlerApiCall} from '@axelor/aos-mobile-core';
 import {searchManagedEmployee as _searchManagedEmployee} from '../api/employee-api';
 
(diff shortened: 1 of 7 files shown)


vhu-axelor (changes requested):
Good job mate!
But don't foget to add `employeeSlice.js` to `asyncFunctions-index.ts` file.

vhu-axelor (approved):

packages/apps/hr/src/features/employeeSlice.js
```diff
@@ -0,0 +1,59 @@
...
+  initialState,
+  extraReducers: builder => {
+    generateInifiniteScrollCases(builder, searchManagedEmployee, {
+      loading: 'loading',
+      moreLoading: 'moreLoading',
+      isListEnd: 'isListEnd',
+      list: 'managedEmployeeList',
+    });
+  },
+});
```
lme-axelor, line 57: You can not use infinite scroll system here with the api request you created as you can not give a page. You asked for all data at once.
Moreover, the objective here is to get the number of employee managed by the used not really the list, so maybe you can use total on request instead of getting the list. 

1. Modify the api request to set a limit on items to fetch otherwise the request can become really heavy (you can also add page mangement if you want to keep the generateInfiniteScrollCases)
2. Ask for total in handlerApiCall (returnTotal or resturnTotalWithData)
3. Modify slice to match the new format of fetched data

vhu-axelor (approved):

PR #304, opened 2023-12-11
fix: Selection container when content is null

gca-axelor:
RM#72919

### DIFF
--- a/packages/ui/src/components/molecules/SelectionContainer/SelectionContainer.tsx
+++ b/packages/ui/src/components/molecules/SelectionContainer/SelectionContainer.tsx
@@ -42,6 +42,8 @@ const SelectionItem = ({
 }: SelectionItemProps) => {
   const Colors = useThemeColor();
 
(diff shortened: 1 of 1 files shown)


gca-axelor:
maybe its a little tricky to test,you can ask me if you want, we can do this together

packages/ui/src/components/molecules/SelectionContainer/SelectionContainer.tsx
```diff
@@ -54,7 +56,7 @@ const SelectionItem = ({
 
   const itemStyles = useMemo(() => getItemStyles(isPicker), [isPicker]);
 
-  return content == null ? null : (
+  return (
```
vhu-axelor, line 59: Remove the condition is enough, you don't have to create `displayContent` const

vhu-axelor (changes requested):

vhu-axelor (approved):

PR #319, opened 2023-12-18
fix : add new slice for myTeamTicket to avoid refresh issue

gca-axelor:
RM#73435

### DIFF
--- a/packages/apps/helpdesk/src/components/templates/MyTeamTicketSearchBar/MyTeamTicketSearchBar.js
+++ b/packages/apps/helpdesk/src/components/templates/MyTeamTicketSearchBar/MyTeamTicketSearchBar.js
@@ -19,9 +19,9 @@
 import React, {useCallback} from 'react';
 import {useDispatch, useSelector, useTranslator} from '@axelor/aos-mobile-core';
 import {AutoCompleteSearch} from '@axelor/aos-mobile-ui';
-import {fetchTickets} from '../../../features/ticketSlice';
+import {fetchMyTeamTickets} from '../../../features/ticketSlice';
 
-const TicketSearchBar = ({
+const MyTeamTicketSearchBar = ({
   placeholderKey = 'Helpdesk_ticket',
   defaultValue = '',
   onChange = () => {},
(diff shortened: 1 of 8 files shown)


vhu-axelor (commented):
~~It works well but I think it will be better to make a new API call when the page is focused again, this would allow you to make fewer modifications and remove the function `fetchMyTeamTickets` which does the same thing as `fetchTickets`~~
Nice work!

packages/apps/helpdesk/src/components/templates/TicketSearchBar/TicketSearchBar.js
```diff
@@ -43,14 +43,15 @@ const TicketSearchBar = ({
...
-      dispatch(
-        fetchTickets({
-          page: page,
-          searchValue: searchValue,
-          userId: team ? null : user.id,
-          userTeam: team ? user.activeTeam : null,
-        }),
-      );
+      team
+        ? dispatch(fetchMyTeamTickets({userTeam: user.activeTeam, page: page}))
```
vhu-axelor, line 47: Add `searchValue: searchValue` here

vhu-axelor (approved):
Nice work!

packages/apps/helpdesk/src/components/templates/TicketSearchBar/TicketSearchBar.js
```diff
@@ -43,14 +43,21 @@ const TicketSearchBar = ({
...
-        }),
-      );
+      team
+        ? dispatch(
+            fetchMyTeamTickets({
+              searchValue: searchValue,
+              userTeam: user.activeTeam,
+              page: page,
+            }),
+          )
```
lme-axelor, line 53: Caution, you are changing the call but not the list used in the searchBar.
I think it could be better to create two searchBar as functionnal use is completly different now.

packages/apps/helpdesk/src/components/templates/MyTeamTicketSearchBar/MyTeamTicketSearchBar.js
```diff
@@ -0,0 +1,78 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React, {useCallback} from 'react';
+import {useDispatch, useSelector, useTranslator} from '@axelor/aos-mobile-core';
+import {AutoCompleteSearch} from '@axelor/aos-mobile-ui';
+import {fetchMyTeamTickets} from '../../../features/ticketSlice';
+
+const MyTeamTicketSearchBar = ({
+  placeholderKey = 'Helpdesk_ticket',
```
vhu-axelor, line 25: In the key ticket has a capital letter -> 'Helpdesk_Ticket'

vhu-axelor (changes requested):
Export `fetchMyTeamTickets` in `asyncFunctions-index.ts`
Maybe rename `fetchTickets` and `TicketSearchBar` because now it matches my tickets and not all tickets

vhu-axelor (approved):

PR #332, opened 2024-01-02
feat: improve ChipSelect component design

vhu-axelor:
RM#73463

### DIFF
--- a/packages/ui/src/components/organisms/ChipSelect/ChipSelect.tsx
+++ b/packages/ui/src/components/organisms/ChipSelect/ChipSelect.tsx
@@ -16,10 +16,13 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import React, {useEffect, useState} from 'react';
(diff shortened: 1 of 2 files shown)


gca-axelor (approved):

packages/ui/src/components/organisms/ChipSelect/ChipSelect.tsx
```diff
@@ -65,6 +68,25 @@ const ChipSelect = ({
...
+    const numberItems = selectionItems.length;
+
+    if (numberItems > 1 && numberItems < 5) {
+      return (
+        (Dimensions.get('window').width -
+          CHIP_CONTAINER_MARGIN * 2 -
+          CHIP_MARGIN * 2 * numberItems) *
+        (1 / numberItems)
+      );
+    }
```
lme-axelor, line 85: If only one chip, component should return null, if more than 5, chip should adjust to content length

gca-axelor (approved):

PR #343, opened 2024-01-08
feat: improve MailMessage filter management

vhu-axelor:
RM#73843


packages/core/src/components/templates/MailMessageView/MailMessageView.js
```diff
@@ -123,7 +123,19 @@ const MailMessageView = ({model, modelId}) => {
   }, [dispatch, model, modelId, comment]);
 
   const filterOnStatus = useCallback(
-    list => filterChip(list, selectedStatus, 'type'),
```
gca-axelor, line 126: remove useless import

PR #347, opened 2024-01-10
feat: add MenuSeparator

gca-axelor:
RM#73773

### DIFF
--- a/packages/core/src/app/Module.ts
+++ b/packages/core/src/app/Module.ts
@@ -28,6 +28,7 @@ interface MenuBase {
   compatibilityAOS?: Compatibility;
   parent?: string;
   order?: number;
+  separator?: boolean;
 }
 
 interface MenuWithSubMenus extends MenuBase {
@@ -37,7 +38,7 @@ interface MenuWithSubMenus extends MenuBase {
 }
 
 interface MenuWithScreen extends MenuBase {
-  screen: string;
+  screen?: string;
 }
 
 interface RootMenuWithScreen extends MenuWithScreen {
--- a/packages/core/src/app/Module.ts
+++ b/packages/core/src/app/Module.ts
@@ -28,7 +28,6 @@ interface MenuBase {
   compatibilityAOS?: Compatibility;
   parent?: string;
   order?: number;
-  separator?: boolean;
 }
 
 interface MenuWithSubMenus extends MenuBase {
@@ -38,14 +37,18 @@ interface MenuWithSubMenus extends MenuBase {
 }
 
 interface MenuWithScreen extends MenuBase {
-  screen?: string;
+  screen: string;
 }
 
 interface RootMenuWithScreen extends MenuWithScreen {
   isDefault?: boolean;
 }
 
-export type Menu = MenuWithSubMenus | RootMenuWithScreen;
+interface MenuSeparator extends MenuBase {
+  separator: true;
+}
+
+export type Menu = MenuWithSubMenus | RootMenuWithScreen | MenuSeparator;
 
 export interface SubMenu extends MenuWithScreen {}
 
--- a/packages/core/src/app/Module.ts
+++ b/packages/core/src/app/Module.ts
@@ -22,7 +22,7 @@ import {FormConfigs} from '../forms/types';
 
 interface MenuBase {
   title: string;
-  icon: string;
+  icon?: string;
   disabled?: boolean;
   hideIf?: (storeState: any) => boolean;
   compatibilityAOS?: Compatibility;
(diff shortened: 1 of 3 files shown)


gca-axelor:
test in any index by add menuentry with propr separator = true


    stock_menu_separator: {
      title: 'Stock_Separator',
      separator: true,
    },

packages/core/src/navigator/drawer/MenuItem.js
```diff
@@ -56,6 +61,7 @@ const MenuItem = ({
       disabled={menuItem.disabled || disabled}
       compatibility={menuItem.compatibilityAOS}
       isActive={isActive}
+      separator={menuItem.separator}
```
vhu-axelor, line 64: Why did you add this props here? MenuItemEntry doesn't seem to use it

packages/core/src/navigator/drawer/MenuSeparator.js
```diff
@@ -0,0 +1,92 @@
...
+  const Colors = useThemeColor();
+
+  const styles = useMemo(() => {
+    return getStyles(Colors);
+  }, [Colors]);
+
+  return (
+    <View style={styles.container}>
+      <View style={styles.childrenContainer}>
+        <View style={[styles.menuItemActive]} />
```
vhu-axelor, line 33: I think you can remove it and thus delete the View with childrenContainer style, don't forget to remove styles of deleted components

packages/core/src/navigator/drawer/MenuSeparator.js
```diff
@@ -0,0 +1,92 @@
...
+        <View style={[styles.menuItemContainer, style]}>
```
vhu-axelor, line 34: If you want to put style props here maybe rename it to make it clear that it allows you to modify this part and not the entire component

packages/core/src/navigator/drawer/MenuSeparator.js
```diff
@@ -0,0 +1,92 @@
...
+      width: '100%',
+    },
+    menuItemActive: {
+      width: 7,
+      height: 32,
+      borderTopRightRadius: 8,
+      borderBottomRightRadius: 8,
+    },
+    menuItemIcon: {
+      marginLeft: 0,
```
vhu-axelor, line 71: I think it's not useful

packages/core/src/navigator/drawer/MenuSeparator.js
```diff
@@ -0,0 +1,92 @@
...
+      <View style={styles.childrenContainer}>
+        <View style={[styles.menuItemActive]} />
+        <View style={[styles.menuItemContainer, style]}>
+          <Icon
+            style={styles.menuItemIcon}
+            name={'chevron-right'}
+            size={iconSize}
+            color={Colors.secondaryColor.background}
+          />
+          <View style={styles.menuItemTextContainer}>
```
vhu-axelor, line 41: Remove it and put style in menuItemContainer

vhu-axelor:
I forgot to put it in my review but the color gives the impression that the entry is disabled

vhu-axelor (approved):

packages/core/src/app/Module.ts
```diff
@@ -37,7 +38,7 @@ interface MenuWithSubMenus extends MenuBase {
 }
 
 interface MenuWithScreen extends MenuBase {
-  screen: string;
+  screen?: string;
```
lme-axelor, line 41: No, that's not logical to have a type name MenuWithScreen and to mark screen as optionnal. You need to create a new type MenuSeparator which should contains only the props taken into account in your separator and then Menu type should be  MenuWithSubMenus | RootMenuWithScreen | MenuSeparator.
Try to factorize the type to avoid duplicated code.
I think separator should take title, hideIf, parent, order and separator props into account

packages/core/src/navigator/drawer/MenuSeparator.js
```diff
@@ -0,0 +1,75 @@
...
+  const styles = useMemo(() => {
+    return getStyles(Colors);
+  }, [Colors]);
+
+  return (
+    <View style={styles.container}>
+      <View style={styles.menuItemContainer}>
+        <Icon
+          style={styles.menuItemIcon}
+          name={'chevron-right'}
```
lme-axelor, line 35:
```
          name='chevron-right'
``` 
is cleaner

packages/core/src/navigator/drawer/MenuSeparator.js
```diff
@@ -0,0 +1,75 @@
...
+          size={iconSize}
+          color={Colors.secondaryColor.background}
+        />
+        <Text style={styles.menuItemTitle}>{title}</Text>
```
lme-axelor, line 39: Do not create a style just for fontSize and text color, you have props for that.
Also, leave default text color for the title

PR #359, opened 2024-01-16
refactor: use DurationInput in DurationFormInput

gca-axelor:
RM#74081


vhu-axelor (changes requested):
The goal here was to remove DurationFormInput and replace it with DurationInput, not to use DurationInput in DurationFormInput

vhu-axelor (approved):
Nice work!

PR #366, opened 2024-01-18
feat: Loader - create Toast management on end request

hel-axelor:
RM#73988


vhu-axelor (changes requested):

packages/core/src/components/external/Loader/use-loader-listener.ts
```diff
@@ -0,0 +1,71 @@
...
+      setLoading(false);
+    }
+  }, [process, onPress]);
+
+  useEffect(() => {
+    if (!start && !loading) {
+      return;
+    }
+
+    handleProcess();
```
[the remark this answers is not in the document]
hel-axelor: Yah, but because I come from java background, we have a programming prettern  **called guard** clause which advises when such conditions are met, the function returns early, preventing the rest of the code from executing.

packages/core/src/components/templates/Loader/use-loader-listener.ts
```diff
@@ -0,0 +1,79 @@
...
+      setStart(false);
+      setLoading(true);
+
+      const response = await process();
+
+      showToastMessage({
+        type: 'success',
+        position: 'top',
+        topOffset: 30,
+        text1: I18n.t('Base_Error'),
```
vhu-axelor, line 50: It's not the right key: `Base_Success`

packages/core/src/components/templates/Loader/use-loader-listener.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  useEffect(() => {
+    if (!start && !loading) {
+      return;
+    }
+
+    executeProcess();
+  }, [start, loading, executeProcess]);
+
+  return useMemo(() => ({loading, listner: () => setStart(true)}), [loading]);
```
vhu-axelor, line 76: You forgot an 'e' in 'listener'

packages/core/src/components/templates/index.js
```diff
@@ -17,6 +17,7 @@
  */
 
 export {default as AttachedFilesView} from './AttachedFilesView/AttachedFilesView';
+export {default as useLoaderListner} from './Loader/use-loader-listener';
```
vhu-axelor, line 20: Add an index file in Loader folder to export this file and the next ones

packages/core/src/i18n/translations/en.json
```diff
@@ -158,6 +158,8 @@
   "Base_Dashboard": "Dashboard",
   "Base_Dashboard_Misconfigured": "This dashboard is misconfigured, please contact your administrator.",
   "Base_Dashboard_RefreshConfig": "Refresh dashboard",
+  "Base_Loader_proccessSuccessMessage": "Process successfully completed.",
+  "Base_Loader_proccessErrorMessage": "An error has occurred!",
```
vhu-axelor, line 162: Caution to capital letter in key: Base_Loader_ProccessSuccessMessage and Base_Loader_ProccessErrorMessage

Also, remove the exclamation point at the end of the error message, it's a bit harsh

packages/core/src/components/templates/Loader/use-loader-listener.ts
```diff
@@ -0,0 +1,79 @@
...
+      setLoading(false);
+    }
+  }, [process, onSuccess, onError, I18n]);
+
+  useEffect(() => {
+    if (!start && !loading) {
+      return;
+    }
+
+    executeProcess();
```
vhu-axelor, line 73: I saw your message and I think that here it only adds lines because the if will be executed in all cases

In addition, for me the guard clause is a condition and the associated action if the condition is met with a return, so here it will be:
```
if (start && !loading) {
  return executeProcess();
}
```

packages/core/src/components/templates/Loader/use-loader-listener.ts
```diff
@@ -0,0 +1,79 @@
...
+        text2: response || I18n.t('Base_Loader_proccessSuccessMessage'),
+        onPress: onSuccess,
+      });
+    } catch (error) {
+      showToastMessage({
+        type: 'error',
+        position: 'top',
+        topOffset: 30,
+        text1: I18n.t('Base_Error'),
+        text2: error.toString() || I18n.t('Base_Loader_proccessErrorMessage'),
```
vhu-axelor, line 60: I don't think you need to add `.toString()`

vhu-axelor (changes requested):

packages/core/src/components/templates/index.js
hel-axelor: I did that in the following ticket

vhu-axelor (changes requested):

PR #367, opened 2024-01-19
feat: add loader wait management popup

hel-axelor:
RM#73990


packages/core/src/components/templates/Loader/LoaderPopup.tsx
```diff
@@ -0,0 +1,118 @@
...
+  Button,
+  Card,
+  Label,
+  Text,
+  useConfig,
+  useThemeColor,
+} from '@axelor/aos-mobile-ui';
+import {ActivityIndicator, Dimensions, StyleSheet, View} from 'react-native';
+import {useTranslator} from '../../../i18n';
+import {useNavigation} from '@react-navigation/native';
```
lme-axelor, line 31: This hook has be rewritten in core package, use this one instead of react-navigation

packages/core/src/components/templates/Loader/LoaderPopup.tsx
```diff
@@ -0,0 +1,118 @@
...
+import {
+  BlockInteractionScreen,
+  Button,
+  Card,
+  Label,
+  Text,
+  useConfig,
+  useThemeColor,
+} from '@axelor/aos-mobile-ui';
+import {ActivityIndicator, Dimensions, StyleSheet, View} from 'react-native';
```
lme-axelor, line 29: Move this import above those of our packages.
In all files, imports follows the same logic : 
1. react packages imports
2. our packages imports
3. current package imports

packages/core/src/components/templates/Loader/LoaderPopup.tsx
```diff
@@ -0,0 +1,118 @@
...
+import {useTranslator} from '../../../i18n';
+import {useNavigation} from '@react-navigation/native';
+
+interface LoaderPopupProps {
+  loading: boolean;
+  timeout: number;
+  onSuccess: () => void;
+  onError: () => void;
+}
```
lme-axelor, line 38: It doesn't match the component props.

packages/apps/stock/src/screens/loader/LoaderScreen.js
```diff
@@ -56,12 +56,14 @@ const LoaderScreen = () => {
     process,
     onSuccess: handleCustomAction,
     onError: () => console.warn('An error has occurred!'),
+    disabled: true,
   });
```
lme-axelor, line 60: The popup component should wrap all the logic, so the use of the hook should be in the popup

packages/core/src/components/templates/Loader/LoaderPopup.tsx
```diff
@@ -0,0 +1,118 @@
...
+  if (!loading || !showPopup) {
+    return null;
+  }
+
+  return (
+    <BlockInteractionScreen hideHeader={true}>
+      <Card style={styles.popupContainer}>
+        <Label
+          type="danger"
+          message={I18n.t('Base_Loading_DoNotCloseTheApp')}
```
lme-axelor, line 74: You did not define any translations.

packages/core/src/components/templates/Loader/LoaderPopup.tsx
```diff
@@ -0,0 +1,118 @@
...
+            size="large"
+            color={Colors.primaryColor.background}
+          />
+          <Text style={styles.loadingLabel}>
+            {I18n.t('Base_Loader_LoadingInProgress')}
+          </Text>
+        </View>
+        <Button
+          iconName="check-lg"
+          title={I18n.t('Base_Loader_NotifyMeWantItIsReady')}
```
lme-axelor, line 88: Translation key can just be Base_Loader_NotifyMe.
You do not need to write all the sentence's words in the key, only the most important to give the global meaning.

packages/core/src/components/templates/Loader/LoaderPopup.tsx
```diff
@@ -0,0 +1,118 @@
...
+          iconName="exclamation-triangle-fill"
+        />
+        <View style={styles.activityIndicatorContainer}>
+          <ActivityIndicator
+            size="large"
+            color={Colors.primaryColor.background}
+          />
+          <Text style={styles.loadingLabel}>
+            {I18n.t('Base_Loader_LoadingInProgress')}
+          </Text>
```
lme-axelor, line 84: Use writing type to make text in bold, you do not need to create a style for that

packages/core/src/components/templates/Loader/LoaderPopup.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const [showPopup, setShowPopup] = useState<boolean>(false);
+
+  const handleGoBack = useCallback(() => {
+    navigation.goBack();
+  }, [navigation]);
+
+  useEffect(() => {
+    if (loading && !showPopup) {
+      setActivityIndicator(true);
+    }
```
lme-axelor, line 55: On first click on the component, you get the right behaviour, that is to say only the activity indicator and then after timeout the details popup.
When you click a second time of the button to trigger the popup, we get directly the detail popup

packages/core/src/components/templates/Loader/LoaderPopup.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+    const timerId = setTimeout(() => {
+      setActivityIndicator(false);
+      setShowPopup(true);
+    }, timeout);
```
lme-axelor, line 60: I think you should create a ref for this timeout to avoid the creation of a new timeout each time one of the props change.

PR #379, opened 2024-01-26
feat: add props on Loader to force user to leave screen

hel-axelor:
RM#73991

### DIFF
--- a/packages/core/src/components/templates/Loader/LoaderPopup.tsx
+++ b/packages/core/src/components/templates/Loader/LoaderPopup.tsx
@@ -63,7 +63,6 @@ const LoaderPopup = ({
   const {processItem, loading} = useLoaderListner(
     {
       disabled,
-      autoLeave,
       process,
       onSuccess,
       onError,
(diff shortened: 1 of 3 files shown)


lme-axelor:
Waiting for #367

packages/core/src/components/templates/Loader/LoaderPopup.tsx
```diff
@@ -61,6 +63,7 @@ const LoaderPopup = ({
   const {processItem, loading} = useLoaderListner(
     {
       disabled,
+      autoLeave,
```
lme-axelor, line 66: I'm not sure, it's useful to add it to the process, you are not using it on the provider. 
What you did in the timeout handler should be sufficient.

packages/core/src/components/templates/Loader/types.ts
```diff
@@ -36,6 +36,7 @@ export enum ProcessStatus {
 
 export type ProcessOption = {
   disabled: boolean;
+  autoLeave: boolean;
```
lme-axelor, line 39: It's not useful here too

packages/core/src/components/templates/Loader/LoaderPopup.tsx
hel-axelor: I added it as process item props in order to save it as history of options related to process.

PR #389, opened 2024-02-05
feat: use Timesheet requests

vhu-axelor:
RM#74837


gca-axelor (commented):
missing action on button in timesheetListScreen
we can't delete timesheet line

vhu-axelor:
> missing action on button in timesheetListScreen we can't delete timesheet line
Thanks, I added the missing actions. For the second point, I can delete timesheet line, maybe you tried to delete one which was referenced by other records.

PR #393, opened 2024-02-08
fix: return graph null when datalist is null

gca-axelor:
RM#74244


packages/ui/src/components/templates/Dashboard/Dashboard.tsx
```diff
@@ -109,15 +109,19 @@ const Dashboard = ({style, lineList}: DashboardProps) => {
...
+        const nbGraphInLine = Math.min(validGraphs.length, MAX_GRAPH_PER_LINE);
+
         const limitedGraphList = line.graphList?.slice(0, MAX_GRAPH_PER_LINE);
 
         return (
           <View style={styles.lineContainer} key={indexLine}>
             {limitedGraphList?.map((graph, indexGraph) => {
+              if (!graph.dataList?.[0]?.length) {
+                return null;
+              }
```
vhu-axelor, line 124: I think you can remove this

vhu-axelor (approved):

PR #399, opened 2024-02-09
fix: can't re-open declare form after one declaration

vhu-axelor:
RM#75193


packages/apps/hr/src/components/templates/TimerListAlert/TimerListAlert.tsx
```diff
@@ -152,6 +152,7 @@ const TimerListAlert = ({
         }),
       );
     }
+    setIsAlertVisible(false);
```
lme-axelor, line 155: I think we should reset the dates/timesheet too
Also, the list view behind is not refresh after validation

PR #413, opened 2024-02-12
feat: add status filter on list screen

vhu-axelor:
RM#75304


packages/apps/quality/src/screens/ControlEntryListScreen.js
```diff
@@ -54,34 +61,44 @@ const ControlEntryListScreen = ({navigation}) => {
...
+                mode="date"
+                popup
+              />
+            </View>
+            <Picker
+              listItems={statusList}
+              title={I18n.t('Base_Status')}
+              onValueChange={setSelectedStatus}
+              labelField="title"
+              valueField="key"
```
lme-axelor, line 101: Please use a ChipSelect instead

PR #438, opened 2024-02-22
feat: add version on SettingsScreen

vhu-axelor:
RM#75791


packages/core/src/app/RootNavigator.js
```diff
@@ -129,7 +122,7 @@ const RootNavigator = ({
             name="SessionManagementScreen"
             component={SessionManagementScreen}
             initialParams={{
-              version,
+              version: appVersion,
```
gca-axelor, line 125: maybe its useless now, you can use appVersion state

gca-axelor (approved):

PR #439, opened 2024-02-22
fix: conflict reducer name 

gca-axelor:
RM#75733


vhu-axelor (changes requested):
You forgot to change the slices name in several components, like in ProductSearchBar, which causes app crash, like when you go to the Active timer screen. Do a global search on VSCode for "state.project" and "state.timer".
Plus, don't forget the changelog.

vhu-axelor (approved):

PR #441, opened 2024-02-26
feat: improve user screen display

vhu-axelor:
RM#75641

### DIFF
--- a/packages/core/src/dashboards/view/DashboardView.tsx
+++ b/packages/core/src/dashboards/view/DashboardView.tsx
@@ -25,7 +25,17 @@ import {ActivityIndicator} from 'react-native';
 import {fetchDashboard} from '../api.helpers';
 import {AOPChart} from '../component';
 
-export const DashboardView = ({dashboardId, hideCardBackground}) => {
+interface DashboardViewProps {
+  dashboardId: number;
+  hideCardBackground?: boolean;
+  dashboardWidth?: number;
+}
+
+export const DashboardView = ({
+  dashboardId,
+  hideCardBackground,
+  dashboardWidth,
+}: DashboardViewProps) => {
   const I18n = useTranslator();
 
   const [dashboard, setDashboard] = useState<any>({});
@@ -99,6 +109,7 @@ export const DashboardView = ({dashboardId, hideCardBackground}) => {
     <Dashboard
       lineList={dashboardData}
       hideCardBackground={hideCardBackground}
+      dashboardWidth={dashboardWidth}
     />
   );
 };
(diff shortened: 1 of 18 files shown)


packages/core/src/dashboards/view/DashboardView.tsx
```diff
@@ -0,0 +1,115 @@
...
+          key: 'refreshConfig',
+          order: 10,
+          showInHeader: false,
+          iconName: 'arrow-repeat',
+          title: I18n.t('Base_Dashboard_RefreshConfig'),
+          onPress: refresh,
+        },
+      ],
+    });
+  }, [I18n, dashboardId, refresh]);
```
lme-axelor, line 79: Add props to enable header action creation or not to avoid it being created on user screen for example as the actionID won't be used, it can be false bu default and mark as true on DashboardScreen component

packages/core/src/dashboards/view/DashboardView.tsx
```diff
@@ -0,0 +1,115 @@
...
+import {useTranslator} from '../../i18n';
+import {createDashboardActionID} from '../display.helpers';
+import {ActivityIndicator} from 'react-native';
+import {fetchDashboard} from '../api.helpers';
+import {AOPChart} from '../component';
+
+interface DashboardViewProps {
+  dashboardId: number;
+  hideCardBackground?: boolean;
+  dashboardWidth?: number;
```
lme-axelor, line 31: chartWidth is more accurate, please change it everywhere

packages/core/src/dashboards/view/DashboardView.tsx
```diff
@@ -0,0 +1,115 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React, {useCallback, useEffect, useMemo, useState} from 'react';
+import {Dashboard, WarningCard} from '@axelor/aos-mobile-ui';
+import {headerActionsProvider} from '../../header';
+import {useTranslator} from '../../i18n';
+import {createDashboardActionID} from '../display.helpers';
+import {ActivityIndicator} from 'react-native';
```
lme-axelor, line 24: Move this under react import please

packages/core/src/dashboards/view/index.ts
```diff
@@ -17,3 +17,4 @@
  */
 
 export {DashboardScreen} from './DashboardScreen';
+export {DashboardView} from './DashboardView';
```
lme-axelor, line 20: Add export in dashboard folder root index to enable use outside of core package

gca-axelor:
very good, but few improvment,
- maybe put number of line at 2 for Shortcut title, 
- look why icon doesnt display interrogation icon when bad name for icon
- In settngsScreen, maybe put a separtor between input and toggle button

PR #460, opened 2024-03-13
feat: add props to hide Card in Dashboard

gca-axelor:
RM#75937

### DIFF
--- a/packages/ui/src/components/templates/Dashboard/Chart/BarChart.tsx
+++ b/packages/ui/src/components/templates/Dashboard/Chart/BarChart.tsx
@@ -17,7 +17,7 @@
  */
 
 import React, {useMemo} from 'react';
-import {StyleSheet, Dimensions} from 'react-native';
+import {StyleSheet, Dimensions, View} from 'react-native';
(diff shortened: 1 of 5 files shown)


packages/ui/src/components/templates/Dashboard/Chart/BarChart.tsx
vhu-axelor, line 1: Instead of creating a function to do the render, I think it would be simpler to do:
```
const Container = hideCardBackground ? View : Card
```
And use it like that:
```
<Container style={[styles.container, {width: _containerWidth}, style]}>
  ...
</Container>
```

packages/ui/src/components/templates/Dashboard/Chart/LineChart.tsx
vhu-axelor, line 1: Same as `BarChart.tsxà

vhu-axelor (commented):
Also, it could be nice to have props `hideCardBackground ` on `DashboardScreen`

vhu-axelor (approved):
Nice work!

PR #471, opened 2024-03-25
feat: add equipmentLine Screen

gca-axelor:
RM#76572


gca-axelor:
depends on 466 need rebase

PR #487, opened 2024-04-03
feat: move TextUnit to ui package

vhu-axelor:
RM#77290


changelogs/unreleased/77290.json
```diff
@@ -0,0 +1,5 @@
+{
+  "title": "TextUnit: move component from hr package to ui package",
+  "type": "feat",
```
lme-axelor, line 3: That's more a refactor than a feat

PR #504, opened 2024-04-10
feat: add possibility to create a new event from a tour line

gca-axelor:
RM#77755


vhu-axelor (changes requested):
When create an event, link it automatically after its creation to the tour line

vhu-axelor (approved):
Nice, works well!

PR #505, opened 2024-04-10
fix:  add default value for currency

gca-axelor:
RM#75564


packages/apps/hr/src/screens/expense/ExpenseLineFormScreen.js
```diff
@@ -171,6 +171,7 @@ const ExpenseLineFormScreen = ({route, navigation}) => {
       totalAmount: 0,
       totalTax: 0,
       distance: 0,
+      currency: user.activeCompany.currency,
```
lme-axelor, line 174: Default value should be given only if currency is managed on ExpenseLine with isMultiCurrencyEnabled config on Mobile settings.
Also, user.activeCompany is not required. So user.activeCompany.currency can throw an error. You need to add a ? as it was done 4 lines above.

vhu-axelor (approved):

PR #507, opened 2024-04-10
fix: add scrollview on drawer to manage small screen

gca-axelor:
RM#77769


vhu-axelor (changes requested):
There is a scroll on text but not for the end of the list (for a long list)

gca-axelor:
hide vertical scrollBar

PR #518, opened 2024-04-18
fix: translation in Event getStatusList function

gca-axelor:
RM#78018


packages/apps/crm/src/types/event-type.ts
```diff
@@ -80,8 +80,8 @@ class EventType {
     Colors: ThemeColors,
     I18n: {t: (key: string) => string},
   ) => {
-    return Object.entries(this.status).map(([key, value]) => ({
```
lme-axelor, line 83: Transform this in Object.values(

PR #519, opened 2024-04-18
fix: question mark when dark mode

gca-axelor:
RM#77717

### DIFF
--- a/changelogs/unreleased/77717.json
+++ b/changelogs/unreleased/77717.json
@@ -1,0 +1,5 @@
+{
+  "title": "BootstrapIconIcon: fix the question mark when dark mode",
+  "type": "fix",
+  "packages": "ui"
+}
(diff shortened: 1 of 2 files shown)


changelogs/unreleased/77717.json
```diff
@@ -0,0 +1,5 @@
+{
+  "title": "BootstrapIconIcon: fix the question mark when dark mode",
```
lme-axelor, line 2:
```
  "title": "BootstrapIcon: define default color to avoid invisible question mark on dark mode",
```

packages/ui/src/components/atoms/BootstrapIcon/BootstrapIcon.tsx
```diff
@@ -55,7 +55,6 @@ const BootstrapIcon = ({name, size, color, style}: BootstrapIconProps) => {
     <Text
       style={{
         fontSize: size / 2,
-        color,
```
lme-axelor, line 58: Instead of doing that, define default color to secondaryColor_dark

PR #529, opened 2024-04-23
fix: refresh issue on active intervention screen

vhu-axelor:
RM#77948


gca-axelor:
I think there is still a issue, when I go to another screen (intervention of the day for example) and I go back to active intervention, nothing is displayed

gca-axelor (approved):

PR #533, opened 2024-04-23
feat: improves Notes card with chevron dropdown logic

gca-axelor:
rm#77951


packages/ui/src/components/molecules/NotesCard/NotesCard.tsx
```diff
@@ -27,16 +30,48 @@ interface NotesCardProps {
...
-      </Card>
+      <View
+        onLayout={event => {
+          const {height} = event.nativeEvent.layout;
+          setChevronHeight(height);
+        }}>
+        <Card
+          style={[
+            styles.note,
+            expanded ? styles.expandedNote : styles.collapsedNote,
```
vhu-axelor, line 57: This condition and these styles are not required, you can only use the style condition in HtmlInput

packages/ui/__tests__/components/molecules/NotesCard.test.js
vhu-axelor, line 1: Test new feature please. I think you can check if there is a chevron icon when the text is big

vhu-axelor (approved):

PR #550, opened 2024-05-02
fix: custom field form do not consider value when field type does not contain attrs

hel-axelor:
RM#78043

### DIFF
--- a/packages/core/src/components/pages/CustomFieldForm/CustomFieldForm.tsx
+++ b/packages/core/src/components/pages/CustomFieldForm/CustomFieldForm.tsx
@@ -44,7 +44,7 @@ interface JsonAction extends Action {
 
 interface CustomFieldFormProps {
   model: string;
-  modelId: string;
+  modelId: number;
   fieldType?: string;
   additionalActions?: JsonAction[];
   readonly?: boolean;
@@ -91,7 +91,7 @@ const CustomFieldForm = ({
   }, [fields, panels, readonly]);
 
   const attrsValues = useMemo(
-    () => (object?.id != modelId ? null : getAttrsValue(object, fieldType)),
+    () => (object?.id !== modelId ? null : getAttrsValue(object, fieldType)),
     [modelId, object, fieldType],
   );
 
--- a/packages/core/src/components/pages/CustomFieldForm/CustomFieldForm.tsx
+++ b/packages/core/src/components/pages/CustomFieldForm/CustomFieldForm.tsx
@@ -91,8 +91,8 @@ const CustomFieldForm = ({
   }, [fields, panels, readonly]);
 
   const attrsValues = useMemo(
-    () => (object?.id !== modelId ? null : getAttrsValue(object)),
-    [modelId, object],
+    () => (object?.id != modelId ? null : getAttrsValue(object, fieldType)),
+    [modelId, object, fieldType],
   );
 
   const _additionalActions: Action[] = useMemo(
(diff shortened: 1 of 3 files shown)


packages/core/src/forms/studio/formula.helpers.ts
```diff
@@ -52,20 +52,16 @@ export const mapStudioFieldsWithFormula = (
...
-  let result = {};
-
-  Object.entries(object)
-    .filter(([key]) => key.toLowerCase().includes('attrs'))
-    .forEach(([_, value]: [string, string]) => {
-      result = {...result, ...JSON.parse(value)};
-    });
-
-  return result;
+  return JSON.parse(object[fieldType]);
```
lme-axelor, line 64: That was not the objective of this ticket. fieldType is not required on CustomFieldForm so it should not be required here too. 
Wanted behaviour : 
* if object is empty : return empty object
* if fieldType is not null and object[fieldType] is not null then return object[fieldType]
* else keep current behaviour to return values of attrs fields

lme-axelor (commented):
You forgot to add a changelog for this fix.
Caution next time to branch name, it does not follow convention : <origin>_<type>_<ticket number>. Ticket number should also be added in PR description to facilitate merge process.

packages/core/src/components/pages/CustomFieldForm/CustomFieldForm.tsx
```diff
@@ -91,8 +91,8 @@ const CustomFieldForm = ({
   }, [fields, panels, readonly]);
 
   const attrsValues = useMemo(
-    () => (object?.id !== modelId ? null : getAttrsValue(object)),
-    [modelId, object],
+    () => (object?.id != modelId ? null : getAttrsValue(object, fieldType)),
```
lme-axelor, line 94: both object?.id and modelId are numbers, it should be compared with !==, and not just !=. Why did you change this ?

packages/core/src/forms/studio/formula.helpers.ts
```diff
@@ -52,11 +52,19 @@ export const mapStudioFieldsWithFormula = (
   }));
 };
 
-export const getAttrsValue = (object: any) => {
+export const getAttrsValue = (object: Object, fieldType: string) => {
```
lme-axelor, line 55: fieldType should not be required, please put a ? in type to make it optional

packages/core/src/forms/studio/formula.helpers.ts
```diff
@@ -52,11 +52,19 @@ export const mapStudioFieldsWithFormula = (
...
   if (isEmpty(object)) {
     return {};
   }
 
+  if (
+    fieldType != null ||
+    object.hasOwnProperty(fieldType) ||
+    typeof object[fieldType] !== 'string'
+  ) {
```
lme-axelor, line 64: This condition is a bit strange.

I think it should be better to replace it with the following. If a fieldType is given then return the parsed value of the object or else a empty object. Also, if you want to use a JSON.parse function, object type **should** be a string but that's not useful to check that here as if a fieldType is given we can suggest that the field will be a JSON field or it will make no sense. 

```
  if (fieldType != null) {
      return object[fieldType] != null ? JSON.parse(object[fieldType]) : {};
  }
```

packages/core/src/components/pages/CustomFieldForm/CustomFieldForm.tsx
hel-axelor: the modelId is not number, it's a string 
https://github.com/axelor/axelor-mobile/pull/550/files/68cfd7053731d690681d3650bdd81ea1a5ea2d2d#diff-104a3b29f081c9e352c7baa5db1fc4450e5aadefb49e54648f2ad3077d3d2a92R47

PR #567, opened 2024-05-22
feat: add SubProjectView in Project module

gca-axelor:
RM#78176

### DIFF
--- a/packages/apps/project/src/components/pages/SubProjectView/SubProjectView.tsx
+++ b/packages/apps/project/src/components/pages/SubProjectView/SubProjectView.tsx
@@ -1,0 +1,141 @@
+/*
+ * Axelor Business Solutions
+ *
+ * Copyright (C) 2024 Axelor (<http://axelor.com>).
+ *
+ * This program is free software: you can redistribute it and/or  modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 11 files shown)


packages/apps/project/src/components/pages/SubProjectView/SubProjectView.tsx
```diff
@@ -0,0 +1,138 @@
...
+                onPress={() => clipboardProvider.copyToClipboard(item?.name)}
+                name={item?.name}
+                code={item?.code}
+                customer={item?.clientPartner}
+                company={item?.company?.name}
+                assignedTo={item?.assignedTo?.fullName}
+                projectStatus={item?.projectStatus}
+                parentProject={item?.parentProject?.fullName}
+                isCopyCard={true}
+              />
```
vhu-axelor, line 113: Too much space between these two components, remove paddingTop of ScrollList

packages/apps/project/src/components/pages/SubProjectView/SubProjectView.tsx
```diff
@@ -0,0 +1,138 @@
...
+            assignedTo={project?.parentProject?.assignedTo?.fullName}
+            projectStatus={project?.parentProject?.projectStatus}
+            parentProject={project?.parentProject?.parentProject?.fullName}
+            isCopyCard={true}
+          />
+        </>
+      )}
+      {subProjectList != null && subProjectList?.length > 0 && (
+        <>
+          <HorizontalRule style={styles.horizontalRule} />
```
vhu-axelor, line 95: Same problem as before, you need to hide if there is no parent project. So, hide it when there is no parent projet or when there is no sub projects. Moreover, add a message (maybe a label), when there is neither

PR #572, opened 2024-05-27
feat: create project task list screen

gca-axelor:
RM#78180


gca-axelor:
need rebase after #561

PR #582, opened 2024-06-04
feat: add TimeView component

gca-axelor:
RM#78178


vhu-axelor (commented):
Don't forget to add hr as a dependency

vhu-axelor (approved):

PR #602, opened 2024-06-18
feat:  add productList Screen on Sales

gca-axelor:
RM#78281

PR #608, opened 2024-06-25
feat: create SaleOrderLineDetailsScreen

vhu-axelor:
RM#80388


vhu-axelor:
Waiting #602 to add ProductCard

PR #610, opened 2024-06-25
fix: manage long title header and scrollView in OperationOrderDetailsScreen

gca-axelor:
RM#80717

### DIFF
--- a/changelogs/unreleased/80717.json
+++ b/changelogs/unreleased/80717.json
@@ -1,0 +1,5 @@
+{
+  "title": "Small screen: manage long title with option and fix scrollView in operation order details screen",
+  "type": "fix",
+  "packages": ["core", "manufacturing"]
+}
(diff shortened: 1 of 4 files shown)


packages/core/src/navigator/drawer/Header.js
```diff
@@ -90,8 +90,8 @@ const Header = ({mainScreen, title, actionID = null, shadedHeader = true}) => {
         ) : (
           <BackIcon tintColor={Colors.primaryColor.background} />
         )}
-        <View>
-          <Text fontSize={20} adjustsFontSizeToFit={true} numberOfLines={1}>
```
vhu-axelor, line 94: Don't remove adjustsFontSizeToFit for "normal" screens

changelogs/unreleased/80717.json
```diff
@@ -0,0 +1,5 @@
+{
+  "title": "Small screen: manage long title with option and fix scrollView in operation order details screen",
+  "type": "fix",
+  "packages": ["core", "manufacturing"]
+}
```
vhu-axelor, line 5: This should be split in two changelogs as changes are not the same at all in the two packages:
- "Operation order: add scroll on details screen",
- "Header: cut long title on small screens"

PR #611, opened 2024-06-25
fix: manage small screens on intervention

gca-axelor:
RM#80726


changelogs/unreleased/80726.json
```diff
@@ -0,0 +1,5 @@
+{
+  "title": "Small screens: add scroll in intervention screen and adjust timer time size depending on screen width",
+  "type": "fix",
+  "packages": ["core", "intervention"]
+}
```
lme-axelor, line 5: This should be splitted in two changelog as changes are not the same at all in the two packages : 
* "Intervention: add scroll on details view",
* "Timer: adjust font size depending on screen width"

PR #614, opened 2024-06-26
fix: BottomBar is cut on iOS due to HeaderBand fix

vhu-axelor:
RM#80956


packages/ui/src/components/templates/BottomBar/BottomBar.tsx
```diff
@@ -17,7 +17,7 @@
  */
 
 import React, {useCallback, useEffect, useMemo, useRef, useState} from 'react';
-import {Dimensions, StyleSheet, View} from 'react-native';
+import {Dimensions, SafeAreaView, StyleSheet, View} from 'react-native';
```
lme-axelor, line 20: Can you try to use react-native-static-safe-area-insets and see if it works please ?

PR #636, opened 2024-07-04
feat: create variant products screen

vhu-axelor:
RM#80373

### DIFF
--- a/packages/apps/sale/src/components/atoms/VariantProductCard/VariantProductCard.tsx
+++ b/packages/apps/sale/src/components/atoms/VariantProductCard/VariantProductCard.tsx
@@ -16,7 +16,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import React, {useCallback} from 'react';
+import React, {useMemo} from 'react';
 import {StyleSheet, View} from 'react-native';
 import {
   useMetafileUri,
@@ -67,7 +67,7 @@ const VariantProductCard = ({
   const formatPrice = usePriceFormat();
   const formatNumber = useDigitFormat();
 
-  const renderAttributes = useCallback(() => {
+  const attributes = useMemo(() => {
     if (
       !Array.isArray(attributesList?.attributes) ||
       attributesList?.attributes.length === 0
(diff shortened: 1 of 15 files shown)


packages/apps/sale/src/components/atoms/VariantProductCard/VariantProductCard.tsx
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    return items?.length > 0 ? {items} : null;
+  }, [
+    attributesList?.attributes,
+    formatNumber,
+    getItemTitle,
+    ProductVariant?.applicationPriceSelect,
+    unit,
+  ]);
```
gca-axelor, line 105: maybe is better to use an useMemo here 

```
  const renderedAttributes = useMemo(() => {
    if (
      !Array.isArray(attributesList?.attributes) ||
      attributesList.attributes.length === 0
    ) {
      return null;
    }

    const items = attributesList?.attributes?.map(attr => {
      const priceExtraText =
        attr.priceExtra > 0
          ? `(${getItemTitle(
              ProductVariant.applicationPriceSelect,
              attr.applicationPriceSelect,
            )} : +${formatNumber(attr.priceExtra)}${unit})`
          : '';

      return {
        displayText: `${attr.attrName} : ${attr.attrValue} ${priceExtraText}`,
        numberOfLines: null,
      };
    });

    return items.length > 0 ? {items} : null;
  }, [
    attributesList?.attributes,
    formatNumber,
    getItemTitle,
    ProductVariant.applicationPriceSelect,
    unit,
  ]);

```

packages/apps/sale/src/components/atoms/VariantProductCard/VariantProductCard.tsx
```diff
@@ -0,0 +1,168 @@
...
+            {
+              displayText: name,
+              isTitle: true,
+            },
+            {
+              displayText: code,
+            },
+          ],
+        }}
+        lowerTexts={renderAttributes()}
```
gca-axelor, line 132: and if do the change upper
        lowerTexts={renderedAttributes}

PR #637, opened 2024-07-05
feat: manage drawing widget on custom form

vhu-axelor:
RM#81364


gca-axelor:
good job, but i have an issue when i put the widget on a form with scrollview, when i draw on it , it makes the screen scrollview moving

lme-axelor:
> good job, but i have an issue when i put the widget on a form with scrollview, when i draw on it , it makes the screen scrollview moving

This issue is not related to this ticket, it will be fixed in another ticket : 81450

PR #644, opened 2024-07-11
fix: display issue for date time inputs on CustomFieldForm

vhu-axelor:
RM#80811


packages/core/src/components/pages/FormView/Components/Field.tsx
```diff
@@ -213,6 +215,7 @@ const Field = ({
             required={isRequired}
             readonly={isReadonly}
             nullable={true}
+            popup={isCustom}
```
lme-axelor, line 218: Do not use isCustom.
Go to manageContentOfModel function in forms/studio/display.helpers.ts and add options attribute to enable popup on component when type is date, datetime or time

PR #663, opened 2024-07-24
feat: add Indicator chart in ChartRender

gca-axelor:
RM#80402


gca-axelor:
http://82542-preview.cloud-sw1.axelor.io/#/ds/menu.admin.app.management/cards

for test

PR #664, opened 2024-07-25
feat: add default interval on period Input

gca-axelor:
RM#80473

### DIFF
--- a/changelogs/unreleased/80473.json
+++ b/changelogs/unreleased/80473.json
@@ -1,5 +1,5 @@
 {
-  "title": "PeriodInput: add interval in between start and end dates with possibility to define a default one in hours with props",
+  "title": "PeriodInput: add interval between start and end dates with possibility to define a default one in hours with props",
   "type": "feat",
   "packages": "core"
 }
--- a/changelogs/unreleased/80473.json
+++ b/changelogs/unreleased/80473.json
@@ -1,0 +1,5 @@
+{
+  "title": "PeriodInput: add defaultInterval to set a default interval between start and end dates",
+  "type": "feat",
+  "packages": "core"
+}
--- a/changelogs/unreleased/80473.json
+++ b/changelogs/unreleased/80473.json
@@ -1,5 +1,5 @@
 {
-  "title": "PeriodInput: add defaultInterval to set a default interval between start and end dates",
+  "title": "PeriodInput: add interval in between start and end dates with possibility to define a default one in hours with props",
   "type": "feat",
   "packages": "core"
 }
--- a/packages/core/src/components/templates/PeriodInput/PeriodInput.tsx
+++ b/packages/core/src/components/templates/PeriodInput/PeriodInput.tsx
@@ -50,7 +50,7 @@ const PeriodInput = ({
   showTitle = true,
   horizontal = true,
   style,
-  defaultInterval,
+  defaultInterval = null,
 }: PeriodInputProps) => {
   const I18n = useTranslator();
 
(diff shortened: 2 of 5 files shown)


packages/core/src/components/templates/PeriodInput/PeriodInput.tsx
```diff
@@ -90,24 +136,22 @@ const PeriodInput = ({
           nullable
           popup={horizontal}
           defaultDate={date}
-          onDateChange={_date => {
-            setDate(_date);
-            dateConfig.onDateChange(_date);
-          }}
+          onDateChange={_date => handleDateChange(mode, _date)}
```
vhu-axelor, line 139: Props mode is use to define isStartDate, instead directly send isStartDate

packages/core/src/components/templates/PeriodInput/PeriodInput.tsx
```diff
@@ -49,12 +50,16 @@ const PeriodInput = ({
...
   style,
+  defaultInterval = null,
 }: PeriodInputProps) => {
   const I18n = useTranslator();
 
   const [startDate, setStartDate] = useState(startDateConfig.date);
   const [endDate, setEndDate] = useState(endDateConfig.date);
   const [isPeriodError, setIsPeriodError] = useState(false);
+  const [interval, setInterval] = useState(
+    defaultInterval !== null ? defaultInterval * 60 * 60 * 1000 : null,
```
vhu-axelor, line 61: Define a global const to transform hours to ms

packages/core/src/components/templates/PeriodInput/PeriodInput.tsx
```diff
@@ -41,6 +41,7 @@ interface PeriodInputProps {
   showTitle?: boolean;
   horizontal?: boolean;
   style?: any;
+  defaultInterval?: number;
```
vhu-axelor, line 44: Maybe add "hours" to props name to make it clearer

changelogs/unreleased/80473.json
```diff
@@ -0,0 +1,5 @@
+{
+  "title": "PeriodInput: add defaultInterval to set a default interval between start and end dates",
```
vhu-axelor, line 2: "PeriodInput: add interval in between start and end dates with possibility to define a default one in hours with props"

vhu-axelor (commented):
When the component receives only one date and the default interval, the second date should be update

packages/core/src/components/templates/PeriodInput/PeriodInput.tsx
```diff
@@ -64,6 +73,16 @@ const PeriodInput = ({
...
   }, [endDateConfig.date]);
 
+  useEffect(() => {
+    if (interval) {
+      if (startDate && !endDate) {
+        setEndDate(new Date(startDate.getTime() + interval));
+      } else if (endDate && !startDate) {
+        setStartDate(new Date(endDate.getTime() - interval));
+      }
+    }
```
lme-axelor, line 83: By doing this this way, the value outside is not updated. You are only updating the visual part but endDate is still empty where the component is used.

packages/core/src/components/templates/PeriodInput/PeriodInput.tsx
```diff
@@ -23,6 +23,8 @@ import {DateInput} from '../../organisms';
 import {getEndOfDay, getStartOfDay} from '../../../utils';
 import {useTranslator} from '../../../i18n';
 
+const hoursToMilliseconds = (hours: number) => hours * 60 * 60 * 1000;
```
lme-axelor, line 26: Maybe we can move this to the time utils file in root utils folder as it is not linked to the component

vhu-axelor (approved):

PR #690, opened 2024-08-28
fix: refresh issue

gca-axelor:
RM#83382

### DIFF
--- a/packages/apps/project/src/components/templates/ActivityListView/ActivityListView.tsx
+++ b/packages/apps/project/src/components/templates/ActivityListView/ActivityListView.tsx
@@ -83,7 +83,7 @@ const ActivityListView = () => {
 
   useEffect(() => {
     if (project?.id) {
-      fetchActivityData(startDate);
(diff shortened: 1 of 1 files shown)


packages/apps/project/src/components/templates/ActivityListView/ActivityListView.tsx
```diff
@@ -94,13 +96,10 @@ const ActivityListView = () => {
...
-    setStartDate(() => {
-      const newDate = new Date();
-      newDate.setDate(newDate.getDate() + 1);
-      return newDate;
-    });
-    setDataList([]);
-    fetchActivityData(new Date()).finally(() => setRefreshing(false));
+    const newDate = new Date();
+    newDate.setDate(newDate.getDate() + 1);
+    setStartDate(newDate);
```
vhu-axelor, line 101: As you set startDate here, it trigger the useEffect which call fetchActivityData without reset at true so the refresh issue is still here

PR #702, opened 2024-09-12
feat: improve project dates display

gca-axelor:
RM#84018

sorry misktake about branch name

### DIFF
--- a/packages/apps/project/src/components/atoms/DatesDisplay/DatesDisplay.tsx
+++ b/packages/apps/project/src/components/atoms/DatesDisplay/DatesDisplay.tsx
@@ -16,10 +16,15 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import React from 'react';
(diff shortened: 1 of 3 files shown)


packages/apps/project/src/components/atoms/DatesDisplay/DatesDisplay.tsx
```diff
@@ -28,16 +28,24 @@ const DatesDisplay = ({
...
 }) => {
+  const I18n = useTranslator();
+
   return (
     <View style={styles.dateContainer}>
-      <DateDisplay date={fromDate} />
+      <View style={styles.dateWrapper}>
+        <Text>{I18n.t('Project_StartedOn')}</Text>
+        <DateDisplay date={fromDate} />
+      </View>
```
lme-axelor, line 38: Create render callback as the structure is the same for both. 
Add a type in args to choose between from and to dates or ask for title and value.

If value is null, it should not render anything
Also center vertically date and title

PR #735, opened 2024-10-01
feat: add reporting view depends on mobileSettings config

gca-axelor:
RM#84017

### DIFF
--- a/packages/apps/project/src/components/pages/ReportingView/ReportingView.tsx
+++ b/packages/apps/project/src/components/pages/ReportingView/ReportingView.tsx
@@ -22,19 +22,16 @@ import {useSelector, useTranslator} from '@axelor/aos-mobile-core';
 import {HeaderContainer, ToggleSwitch} from '@axelor/aos-mobile-ui';
 import {ProjectHeader} from '../../molecules';
 import {ActivityListView, ReportingDetailsView} from '../../templates';
-import {getReportingConfiguration, ReportingType} from '../../../utils';
+import {ReportingType} from '../../../types';
+import {useReportingConfiguration} from '../../../hooks/use-reporting-configuration';
 
 const ReportingView = () => {
   const I18n = useTranslator();
 
   const {project} = useSelector((state: any) => state.project_project);
-  const {mobileSettings} = useSelector((state: any) => state.appConfig);
 
-  const allowedReportingTypes = mobileSettings?.reportingTypesToDisplay || [];
-  const {isNoneOnly, showActivities, showReporting} = getReportingConfiguration(
-    allowedReportingTypes,
-    project,
-  );
+  const {isNoneOnly, showActivities, showReporting} =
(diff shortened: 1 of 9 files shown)


packages/apps/project/src/utils/reportingUtils.ts
vhu-axelor, line 1: Don't hard code reporting types, use a type file instead

vhu-axelor (approved):

packages/apps/project/src/utils/reportingUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+    allowedReportingTypes.includes(ReportingType.indicator) &&
+    project?.isBusinessProject;
+
+  return {
+    isNoneOnly,
+    showActivities,
+    showReporting,
+    showReportingOrActivities: showActivities || showReporting,
+  };
+};
```
lme-axelor, line 42: Create hook instead of utils function. You will be able to access mobileSettings with useSelector and return useMemo values.

packages/apps/project/src/utils/reportingUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+  );
+  const showReporting =
+    allowedReportingTypes.includes(ReportingType.indicator) &&
+    project?.isBusinessProject;
+
+  return {
+    isNoneOnly,
+    showActivities,
+    showReporting,
+    showReportingOrActivities: showActivities || showReporting,
```
lme-axelor, line 40: isNoneOnly and showReportingOrActivities is a bit redundant. 
Maybe you could first evaluate showActivities and showIndicators and return a noReporting value with `!showActivities && !noReporting`.

If configuration is defined on none, the API will return 'indicators' and 'activities' values, so I think it is not necessary to check none value.

packages/apps/project/src/components/pages/ReportingView/ReportingView.tsx
```diff
@@ -48,26 +57,26 @@ const ReportingView = () => {
         fixedItems={
           <>
             <ProjectHeader />
-            {project?.isBusinessProject && (
+            {project?.isBusinessProject && showReporting && showActivities && (
```
lme-axelor, line 60: As business project is already checked in hook function, you can only check showReporting && showActivities

packages/apps/project/src/components/pages/ReportingView/ReportingView.tsx
```diff
@@ -48,26 +57,26 @@ const ReportingView = () => {
...
               />
             )}
           </>
         }
       />
-      {mode === modes.activities ? (
+      {mode === ReportingType.activities && showActivities ? (
         <ActivityListView />
       ) : (
-        <ReportingDetailsView />
```
lme-axelor, line 70: Only check current mode as you create a useEffect to switch mode depending on configuration

PR #736, opened 2024-10-02
feat: add missing quantity tag on stock move line

vhu-axelor:
RM#80182

### DIFF
--- a/packages/apps/stock/src/components/organisms/StockMoveHeader/StockMoveHeader.tsx
+++ b/packages/apps/stock/src/components/organisms/StockMoveHeader/StockMoveHeader.tsx
@@ -58,7 +58,6 @@ const StockMoveHeader = ({
   const [checkQtyObject, setCheckQtyObject] = useState(null);
 
   useEffect(() => {
-    console.log(stockMoveLineId);
     if (stockMoveLineId) {
       checkQuantityApi({stockMoveLineId})
         .then(response => {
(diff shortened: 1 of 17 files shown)


packages/apps/stock/src/components/templates/customerDelivery/CustomerDeliveryLineCard/CustomerDeliveryLineCard.tsx
```diff
@@ -75,15 +79,25 @@ const CustomerDeliveryLineCard = ({
...
 
+  useEffect(() => {
+    checkQuantityApi({stockMoveLineId})
+      .then(response => {
+        if (response?.data?.object) {
+          setCheckQtyObject(response?.data?.object);
+        }
+      })
+      .catch(() => setCheckQtyObject(null));
+  }, [stockMoveLineId]);
```
lme-axelor, line 94: Can we imagine to create a hook to manage the process, as it is the same thing on multiple screen/components

packages/apps/stock/src/components/organisms/StockMoveHeader/StockMoveHeader.tsx
```diff
@@ -65,43 +89,43 @@ const StockMoveHeader = ({
...
+            />
+          )}
           <Badge
             color={getItemColor(StockMove?.availableStatusSelect, availability)}
-            title={getItemTitle(StockMove?.availableStatusSelect, availability)}
+            title={
+              checkQtyObject?.availability ??
+              getItemTitle(StockMove?.availableStatusSelect, availability)
+            }
           />
```
lme-axelor, line 110: This need to be hidden when availableStatusSelect is null or 0 (example on Supplier arrival)

PR #777, opened 2024-10-28
feat: enable login page override

gca-axelor:
RM#86091


gca-axelor:
import this file on app props 
      customLoginPage={CustomLogin}
```
import React, {useEffect} from 'react';
import {Text} from '@axelor/aos-mobile-ui';
import {
  useDispatch,
  useSelector,
  updateAuthState,
} from '@axelor/aos-mobile-core';

const CustomLogin = () => {
  const {token} = useSelector(state => state.auth);
  const dispatch = useDispatch();

  useEffect(() => {
    dispatch(updateAuthState({key: 'token', value: 'newToken'}));
  }, [dispatch]);

  console.log(token);

  return <Text>Custom</Text>;
};

export default CustomLogin;
```

PR #792, opened 2024-11-06
fix: merge configuration

gca-axelor:
RM#86276


gca-axelor:
you can test this PR on wip with user testPerm , password Zizou1998! on stockMoveLine perm (r and w)

PR #815, opened 2024-11-28
feat: add possibility to define default value for parent filter on TreeView

vhu-axelor:
RM#86491


vhu-axelor:
PR #803 need to be merged before

PR #819, opened 2024-12-02
feat: add RequestListView

gca-axelor:
RM#86192


packages/apps/purchase/src/api/supplier-api.js
```diff
@@ -0,0 +1,50 @@
...
+
+  return criterias;
+};
+
+export async function searchSupplier({page = 0, searchValue}) {
+  return createStandardSearch({
+    model: 'com.axelor.apps.base.db.Partner',
+    criteria: createSupplierCriteria(searchValue),
+    fieldKey: 'purchase_supplier',
+    sortKey: 'purchase_supplier',
```
vhu-axelor, line 47: Declare it or remove it

packages/apps/purchase/src/models/typeObjects.ts
```diff
@@ -0,0 +1,67 @@
...
+          {
+            key: 'Requested',
+            value: 2,
+            title: 'Purchase_Status_Requested',
+            color: 'progressColor',
+          },
+          {
+            key: 'Accepted',
+            value: 3,
+            title: 'Purchase_Status_Planned',
```
vhu-axelor, line 42: Change the title (and trads behind) to match the key

vhu-axelor (approved):

packages/apps/purchase/src/api/supplier-api.js
lme-axelor: That's not true. As a good practice, those keys should be always present to enable client projects to override the value. If it's not present, then the client project will need to rewrite the whole API/slice only to modify sorting.

So if we don't need it in standard, then no need to declare it but it should always be put in the API call so some other package can easily declare it.

packages/apps/purchase/src/i18n/fr.json
```diff
@@ -1,3 +1,13 @@
 {
-  "Purchase_Purchase": "Achats"
+  "Purchase_Purchase": "Achats",
+  "Purchase_InternalRequests": "Demandes internes",
+  "Purchase_Status_Draft": "Brouillon",
+  "Purchase_Status_Requested": "Demandé",
+  "Purchase_Status_Accepted": "Accepté",
+  "Purchase_Status_Purchased": "Acheté",
+  "Purchase_Status_Refused": "Refusé",
+  "Purchase_Status_Canceled": "Annulé",
```
lme-axelor, line 9: I think we should be an ending 'e' on this statuses as "Demande" is a feminine name

packages/apps/purchase/src/models/sortFields.ts
```diff
@@ -0,0 +1,23 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import {SortFields} from '@axelor/aos-mobile-core';
+
+export const purchase_sortFields: SortFields = {
+  purchase_purchaseRequest: ['statusSelect'],
```
lme-axelor, line 22: sort them by purchaseRequestSeq too to get the latest request as the top for each status

packages/apps/purchase/src/models/typeObjects.ts
```diff
@@ -0,0 +1,67 @@
...
+            key: 'Refused',
+            value: 5,
+            title: 'Purchase_Status_Refused',
+            color: 'cautionColor',
+          },
+          {
+            key: 'Canceled',
+            value: 6,
+            title: 'Purchase_Status_Canceled',
+            color: 'errorColor',
```
lme-axelor, line 61: Inverse the color of those statuses

packages/apps/purchase/src/screens/RequestListScreen.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  const [selectedStatus, setSelectedStatus] = useState([]);
+  const [supplier, setSupplier] = useState();
+
+  const statusList = useMemo(() => {
+    const statusToDisplay = [
+      PurchaseRequest?.statusSelect.Draft,
+      PurchaseRequest?.statusSelect.Requested,
+      PurchaseRequest?.statusSelect.Accepted,
+    ];
```
lme-axelor, line 53: I think we should either show all statuses in the ChipSelect or only display the concerned statuses in the list.
Currently, it's not really consistent. Let's check with MCH when he comes back.

PR #820, opened 2024-12-02
feat: create DMSFile form screen

vhu-axelor:
RM#87125

### DIFF
--- a/packages/apps/dms/src/components/organisms/ParentDirectorySearchBar/ParentDirectorySearchBar.tsx
+++ b/packages/apps/dms/src/components/organisms/ParentDirectorySearchBar/ParentDirectorySearchBar.tsx
@@ -1,0 +1,94 @@
+/*
+ * Axelor Business Solutions
+ *
+ * Copyright (C) 2024 Axelor (<http://axelor.com>).
+ *
+ * This program is free software: you can redistribute it and/or  modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import React, {useCallback, useMemo} from 'react';
+import {useDispatch, useSelector, useTranslator} from '@axelor/aos-mobile-core';
+import {AutoCompleteSearch} from '@axelor/aos-mobile-ui';
+import {searchDirectory} from '../../../features/documentSlice';
+
+interface ParentDirectorySearchBarProps {
+  style?: any;
+  title?: string;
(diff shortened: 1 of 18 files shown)


packages/apps/dms/src/models/forms.ts
gca-axelor, line 1: I think you can put isDirectory as readonly when you are not in creation mode to avoid losing child directory or files

packages/apps/dms/src/models/forms.ts
```diff
@@ -0,0 +1,69 @@
...
+      },
+      isDirectory: {
+        titleKey: 'Dms_Folder',
+        type: 'boolean',
+        widget: 'checkbox',
+        hideIf: ({objectState}) => objectState.id,
+        options: {
+          iconSize: 20,
+        },
+      },
```
lme-axelor, line 56: Maybe we can improve a bit the style of the form, use panel to make name and directory as a row ? 
Remove bold on the checkbox title ? That's not really nice for the moment 😅

packages/apps/dms/src/screens/DocumentFormScreen.tsx
```diff
@@ -0,0 +1,103 @@
...
+      }
+
+      if (parentId === mobileSettings.defaultDmsRoot?.id) {
+        _document.parent = mobileSettings.defaultDmsRoot;
+      }
+
+      const sliceFunction = isCreation ? createDocument : updateDocument;
+      dispatch((sliceFunction as any)({document: _document}));
+
+      navigation.navigate('AllDocumentsScreen');
```
lme-axelor, line 68: Just pop here, we might come from other screens than AllDocumentsScreen

packages/apps/dms/src/screens/index.ts
```diff
@@ -35,7 +37,12 @@ export default {
       shadedHeader: false,
     },
   },
+  DocumentFormScreen: {
+    title: 'Dms_Document',
+    component: DocumentFormScreen,
```
lme-axelor, line 42: I think we can mark this screen as usable on shortcuts

packages/apps/dms/src/components/organisms/ParentDirectorySearchBar/ParentDirectorySearchBar.tsx
```diff
@@ -0,0 +1,94 @@
...
+    loadingDirectory,
+    moreLoadingDirectory,
+    isListEndDirectory,
+    directoryList,
+  } = useSelector((state: any) => state.dms_document);
+
+  const extendedDirectoryList = useMemo(
+    () => [
+      {
+        ...(user.dmsRoot ?? mobileSettings.defaultDmsRoot),
```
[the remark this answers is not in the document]
vhu-axelor: If both are undefined the object will only have the props fileName which is what I want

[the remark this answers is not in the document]
lme-axelor: I think if both are undefined, it will crash as you can not use ... operator on a undefined object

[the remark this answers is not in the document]
vhu-axelor: No, it's not a problem as JS ignore the destruction if the value is undefined

PR #827, opened 2024-12-04
fix: filter issue with chips on DocumentList

vhu-axelor:
RM#87963


packages/apps/dms/src/components/templates/DocumentList/DocumentList.tsx
```diff
@@ -59,9 +59,8 @@ const DocumentList = ({defaultParent}: DocumentListProps) => {
   const sliceFunctionData = useMemo(
     () => ({
       authorId: author?.id,
-      extensions: selectedExtensions.map(_extension => _extension.key),
     }),
-    [author?.id, selectedExtensions],
```
lme-axelor, line 64: This should not be removed or the root elements won't be filtered

PR #836, opened 2024-12-11
fix: contact and partner mutually dependent

gca-axelor:
RM#88244

### DIFF
--- a/packages/apps/crm/src/models/forms.ts
+++ b/packages/apps/crm/src/models/forms.ts
@@ -302,6 +302,11 @@ export const crm_formsRegister: FormConfigs = {
         widget: 'custom',
         customComponent: ClientProspectSearchBar,
         required: true,
+        dependsOn: {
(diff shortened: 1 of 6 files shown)


vhu-axelor (approved):

packages/apps/crm/src/models/forms.ts
```diff
@@ -302,6 +302,11 @@ export const crm_formsRegister: FormConfigs = {
         widget: 'custom',
         customComponent: ClientProspectSearchBar,
         required: true,
+        dependsOn: {
+          contact: ({newValue}) => {
+            return newValue?.mainPartner;
+          },
+        },
```
lme-axelor, line 309: Should be apply only the contact is not null.
If we unselect the contact, it should not clear the partner too

Also, we need to manage the partner update too. If we clear or change the partner to a value different from contact.mainPartner, it should clear the contact

vhu-axelor (approved):

PR #850, opened 2024-12-17
refactor: manage company in data fetching in manufacturing

vhu-axelor:
RM#88459


lme-axelor (commented):
You need to add the company management on every calls that is accessible directly from the menu so on the OperationOrder object too using manufOrder.company.
Also you need to add it on every search on ManufacturingOrder object as it's on this object that the company is managed so add it on the linked and children manuf order calls

PR #858, opened 2024-12-19
refactor: manage company in data fetching in project

gca-axelor:
RM#88460


lme-axelor (commented):
I think we need to add the filter on the ProjectTask object too through the linked project as the tasks are accessible directly from the menu

And missing changelog

PR #860, opened 2024-12-30
feat: add option to ignore action api

vhu-axelor:
RM#86533

### DIFF
--- a/packages/core/src/apiProviders/Action/AosActionApi.ts
+++ b/packages/core/src/apiProviders/Action/AosActionApi.ts
@@ -39,6 +39,10 @@ export class AosActionApi implements ActionApi {
   }
 
   send(request: ActionRequest): Promise<void> {
+    if (request.ignoreRequest) {
+      return null;
+    }
+
     return axios[request.method](request.url, request.body);
   }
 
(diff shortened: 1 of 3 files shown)


packages/core/src/apiProviders/Action/AosActionApi.ts
```diff
@@ -39,6 +39,10 @@ export class AosActionApi implements ActionApi {
   }
 
   send(request: ActionRequest): Promise<void> {
+    if (request.ignoreRequest) {
+      return null;
+    }
```
lme-axelor, line 44: Please remove this, in AosActionApi we won't implement the ignoreRequest option else there is no interest in doing the request initially 😅
This option is especially designed for providers who set a delay before executing requests, and to queries that are only of interest at the present time.

vhu-axelor: I thought it might be useful to have this option if we need to condition the execution of a query based on its parameters. That way, instead of doing it every time we make the API call, we do it directly in it to avoid code duplication

lme-axelor: i understand your idea but if something like this is necessary, it's need to be done inside the api function before calling the provider, this option is generic to all calls and depends on provider logic.

PR #877, opened 2025-01-22
chore: update react native to 0.74

gca-axelor:
RM#89397


ios/Podfile.lock
lme-axelor, line 1: This will need to be updated @vhu-axelor before merging

PR #892, opened 2025-02-14
feat: add API action on request creation

vhu-axelor:
RM#89562

### DIFF
--- a/packages/apps/purchase/src/components/templates/RequestCreationButtons/RequestCreationButtons.tsx
+++ b/packages/apps/purchase/src/components/templates/RequestCreationButtons/RequestCreationButtons.tsx
@@ -25,7 +25,6 @@ import {RequestCreation} from '../../../types';
 
 interface RequestCreationButtonsProps {
   step: number;
-  setStep: (step: number) => void;
   lines: any[];
   isEditionMode?: boolean;
   disabled?: boolean;
@@ -37,7 +36,6 @@ interface RequestCreationButtonsProps {
 
 const RequestCreationButtons = ({
   step,
-  setStep,
   lines,
   isEditionMode = false,
   disabled = false,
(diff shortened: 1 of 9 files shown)


packages/apps/purchase/src/screens/RequestCreationScreen.tsx
```diff
@@ -72,7 +79,7 @@ const RequestCreationScreen = () => {
   const handleProductChange = useCallback(
     (_value: any) => {
       if (_value == null) {
-        handleReset(RequestCreation.step.addLine);
+        newLine != null && handleReset(RequestCreation.step.addLine);
```
lme-axelor, line 82: Not a good practice to add as a dependency when the setter is used, can you check for another solution please ?

packages/apps/purchase/src/components/templates/RequestCreationButtons/RequestCreationButtons.tsx
```diff
@@ -54,7 +63,31 @@ const RequestCreationButtons = ({
     setStep(RequestCreation.step.finish);
   };
 
-  const handleRealizePress = useCallback(() => {}, []);
+  const handleRealizePress = useCallback(() => {
```
lme-axelor, line 66: Also can you check step management please, I think the finish step is not useful as there is no end information to set. 
Currently we need to click twice on the finish button

lme-axelor: Still an issue on step management, currently I need to add the line to be able to finish the request but the button is always visible so two options : 
* hide the realize button when adding a new line so it clear to user that he need to add the line before validating
* always display the realize button but it should do the line addition before creating the request. The disabled condition will be the same as the add button

I think the first option is simplier but I'm okay with both if you prefer the second

PR #906, opened 2025-02-26
fix: crash when no management of parent from outside of tree view

vhu-axelor:
RM#91546


changelogs/unreleased/91546.json
```diff
@@ -0,0 +1,5 @@
+{
+  "title": "SearchTreeView: resolve issue when onParentChange not provided",
+  "type": "fix",
+  "packages": "core"
```
lme-axelor, line 4: No changelog here, it's coming from the 8.3

PR #915, opened 2025-03-03
refactor: manage company in leave api

gca-axelor:
RM#91557


vhu-axelor (changes requested):
In `leaveSlice.js` file `fetchLeave` and `fetchLeaveToValidate` are used in many functions, add companyId in these functions too

vhu-axelor (approved):

PR #922, opened 2025-03-11
fix: add showClearButton on DMS chipSelect

gca-axelor:
RM#92153


packages/apps/dms/src/components/templates/DocumentList/DocumentList.tsx
```diff
@@ -179,6 +179,7 @@ const DocumentList = ({
             mode="multi"
             selectionItems={File.getFileExtensionList(Colors)}
             onChangeValue={setSelectedExtensions}
+            showClearButton
```
lme-axelor, line 182: It was asked in ticket to add it when File.getFileExtentionList is called so here and in MyFavoriteDocumentsScreen

PR #933, opened 2025-03-17
Fix: POST to //callback during log-in.

mjkaye:
When I was trying to log in using the mobile app, it was trying to POST to `//callback`, which failed and resulted in a `400` status code. I believe this may have something to do with the fact that I serve AOS from the root of my domain, and that it may be an issue for others. This commit ensures that the POST URI is correctly formatted.

### DIFF
--- a/packages/core/src/api/login-api.ts
+++ b/packages/core/src/api/login-api.ts
@@ -75,8 +75,10 @@ export async function loginApi(
   requestInterceptorId: number;
   responseInterceptorId: number;
 }> {
+  const normalizedUrl = url.endsWith('/') ? url.slice(0, -1) : url;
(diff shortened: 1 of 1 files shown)


lme-axelor:
Thanks for raising this anomaly, it might indeed be due to your server configuration as we never detected any issue on this behavior.

The base url given to that function is always formatted thanks to the [formatURL function](https://github.com/axelor/axelor-mobile/blob/main/packages/core/src/utils/formatters.ts#L100). This function ensures that the base url always end with a slash symbol. So no need to normalize the url argument more. 
To fix your issue, we are going to remove the starting slash in the loginPath variable which is indeed not needed. This fix will be available in next patch version.

PR #934, opened 2025-03-17
fix: display issue line card

vhu-axelor:
RM#92436

### DIFF
--- a/packages/ui/src/components/molecules/LabelText/LabelText.tsx
+++ b/packages/ui/src/components/molecules/LabelText/LabelText.tsx
@@ -46,6 +46,22 @@ const LabelText = ({
   iconName = null,
   onlyOneLine = false,
 }: LabelTextProps) => {
+  const renderValue = () => {
+    if (checkNullString(value?.toString())) {
+      return null;
+    }
+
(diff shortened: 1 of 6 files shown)


gca-axelor (approved):

packages/ui/src/components/molecules/LabelText/LabelText.tsx
```diff
@@ -56,20 +72,15 @@ const LabelText = ({
...
+      ) : (
         <Text
-          style={[styles.txtDetails, textStyle]}
+          style={textStyle}
           fontSize={textSize}
           numberOfLines={onlyOneLine ? 1 : null}>
-          {value}
+          {`${title} `}
+          {renderValue()}
         </Text>
```
lme-axelor, line 84: Can't we simplify a bit ? 

Something like that : 
```
      <Text
        style={textStyle}
        fontSize={textSize}
        numberOfLines={onlyOneLine ? 1 : null}>
        {!checkNullString(title?.toString()) && `${title} `}
        {renderValue()}
      </Text>
```

With that solution, the number of lines might not be necessary in the value render

PR #936, opened 2025-03-18
refactor: use AOP filter on CRM list views

gca-axelor:
RM#92478

### DIFF
--- a/packages/apps/crm/src/api/leads-api.js
+++ b/packages/apps/crm/src/api/leads-api.js
@@ -74,19 +74,7 @@ export async function searchLeads({
     sortKey: 'crm_lead',
     page,
     provider: 'model',
-    ...(filterDomain && {
-      domains: [
-        {
-          type: filterDomain.type,
-          name: filterDomain.name,
-          domain: filterDomain.domain,
-          title: filterDomain.title,
-        },
-      ],
-      domainContext: {
-        _model: 'com.axelor.apps.crm.db.Lead',
-      },
-    }),
+    domains: filterDomain,
   });
 }
 
--- a/packages/apps/crm/src/api/leads-api.js
+++ b/packages/apps/crm/src/api/leads-api.js
@@ -66,7 +66,6 @@ export async function searchLeads({
   companyId,
   filterDomain,
 }) {
-  console.log('filterDomain', filterDomain);
   return createStandardSearch({
     model: 'com.axelor.apps.crm.db.Lead',
     companyId,
--- a/packages/apps/crm/src/api/leads-api.js
+++ b/packages/apps/crm/src/api/leads-api.js
@@ -64,7 +64,9 @@ export async function searchLeads({
   assigned,
   statusList,
   companyId,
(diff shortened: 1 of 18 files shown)


packages/core/src/apiProviders/Standard/requests.helper.ts
```diff
@@ -120,6 +122,18 @@ class RequestBuilder {
         data._domainContext = domainContext;
       }
     }
+    if (domains != null && domains !== '') {
+      if (data._domains != null) {
+        data._domains = `(${data._domains}) AND (${domains})`;
+        data._domainContext = {
+          ...data._domainContext,
+          ...domainContext,
+        };
```
lme-axelor, line 131: This will never occur as you did not set anything in data before this condition. Also domains is not a string like the classic domain attribut so it would have not work

packages/core/src/apiProviders/Standard/requests.helper.ts
```diff
@@ -120,6 +122,18 @@ class RequestBuilder {
         data._domainContext = domainContext;
       }
     }
+    if (domains != null && domains !== '') {
```
lme-axelor, line 125: domains is not a string but an array

packages/apps/crm/src/api/leads-api.js
```diff
@@ -73,6 +75,19 @@ export async function searchLeads({
...
+          type: filterDomain.type,
+          name: filterDomain.name,
+          domain: filterDomain.domain,
+          title: filterDomain.title,
+        },
+      ],
+      domainContext: {
+        _model: 'com.axelor.apps.crm.db.Lead',
+      },
+    }),
```
lme-axelor, line 90: This will be done for each request configured for aop filter so we need to create a tool for that. I think it can be integrated directly in the createStandardSearch function.
So two things to do : create a function to help the domain creation based on model and filter object and use that tool in createStandardSearch to create the domains criteria

packages/core/src/apiProviders/Standard/requests.helper.ts
```diff
@@ -120,6 +122,18 @@ class RequestBuilder {
...
+      if (data._domains != null) {
+        data._domains = `(${data._domains}) AND (${domains})`;
+        data._domainContext = {
+          ...data._domainContext,
+          ...domainContext,
+        };
+      } else {
+        data._domains = domains;
+        data._domainContext = domainContext;
```
lme-axelor, line 134: Caution here, data._domainContext can be not empty, you need to merge existing context with args value and newly created tool result

lme-axelor (commented):
Please modify current feedbacks on **lead list view only** and once this is validated, we will continue the refactor in CRM package and others. We need to validate the logic before doing the same everywhere

lme-axelor:
You can continue the refactor on the screens mentioned on the associated ticket. Do not forget to add a changelog.
Something like "List views: add AOP filter management"

vhu-axelor (approved):

PR #937, opened 2025-03-18
fix: add missing header action to create attached file

vhu-axelor:
RM#92414


packages/apps/dms/src/screens/AttachedFilesScreen.tsx
```diff
@@ -56,12 +64,36 @@ const AttachedFilesScreen = ({navigation, route}) => {
...
+    });
+  }, [
+    Colors.primaryColor.background,
+    I18n,
+    canCreate,
+    mobileSettings,
+    navigation,
+    options,
+    parent,
+  ]);
```
lme-axelor, line 96: Please move that to the DMS headers hook
Only the screen title should be left here.

vhu-axelor: I need to access `parent` state so I think I can't move that to the DMS headers hook

lme-axelor: You are right, I didn't see it the first time

packages/apps/dms/src/screens/AttachedFilesScreen.tsx
```diff
@@ -56,12 +64,36 @@ const AttachedFilesScreen = ({navigation, route}) => {
...
+          iconName: 'plus-lg',
+          title: I18n.t('Dms_NewDocument'),
+          iconColor: Colors.primaryColor.background,
+          hideIf:
+            !canCreate ||
+            !mobileSettings?.isFolderCreationAllowed ||
+            !mobileSettings?.isFileCreationAllowed,
+          onPress: () =>
+            navigation.navigate('DocumentFormScreen', {
+              parent,
```
lme-axelor, line 82: Use the model & modelId information instead like it's done for the alert confirm button

packages/apps/dms/src/screens/AttachedFilesScreen.tsx
```diff
@@ -56,12 +64,36 @@ const AttachedFilesScreen = ({navigation, route}) => {
...
+        {
+          key: 'newDocument',
+          order: 10,
+          iconName: 'plus-lg',
+          title: I18n.t('Dms_NewDocument'),
+          iconColor: Colors.primaryColor.background,
+          hideIf:
+            !canCreate ||
+            !mobileSettings?.isFolderCreationAllowed ||
+            !mobileSettings?.isFileCreationAllowed,
```
lme-axelor, line 79: Condition is not right. It should be hidden if both configurations are false. If it's only file or only folder, user should still be able to access the form

PR #938, opened 2025-03-19
fix: date not save on click outside on DateInput

vhu-axelor:
RM#92548


gca-axelor (approved):

packages/core/src/components/organisms/DateInput/DateInput.tsx
```diff
@@ -184,9 +184,10 @@ const DateInput = ({
 
   useEffect(() => {
     if (clickOutside === OUTSIDE_INDICATOR && pickerIsOpen) {
+      onDateChange(selectedDate);
```
lme-axelor, line 187: We are going to reset the old value on click outside instead
We will see in a later version to improve the component behavior

PR #939, opened 2025-03-19
 fix: crash when trying to add an event from the planning

vhu-axelor:
RM#92549

### DIFF
--- a/packages/apps/crm/src/screens/event/EventFormScreen.js
+++ b/packages/apps/crm/src/screens/event/EventFormScreen.js
@@ -39,9 +39,15 @@ const EventFormScreen = ({navigation, route}) => {
   const {user} = useSelector(state => state.user);
 
   const creationDefaultValue = useMemo(() => {
-    const _defaultStartDate = new Date();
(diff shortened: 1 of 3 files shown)


packages/apps/crm/src/screens/event/EventFormScreen.js
```diff
@@ -39,9 +39,15 @@ const EventFormScreen = ({navigation, route}) => {
...
-
+    const today = new Date();
+    const baseDate = eventPlanningDate
+      ? new Date(eventPlanningDate).setHours(
+          today.getHours(),
+          today.getMinutes(),
+        )
+      : today;
+    const _defaultStartDate = new Date(baseDate);
+    const _defaultEndDate = new Date(baseDate);
```
gca-axelor, line 50: maybe i miising something, but it seems to have an issue when you create an event, the endDateTime is corrctly set vsually but yhis value is actually the same than the startdate

gca-axelor: The problem does not come from this pr, a ticket has been created to solve it.

PR #942, opened 2025-03-24
fix: improve navigation after line validation

gca-axelor:
RM#92404

### DIFF
--- a/changelogs/unreleased/92404.json
+++ b/changelogs/unreleased/92404.json
@@ -1,5 +1,5 @@
 {
-  "title": "Navigation stock move line: improve navigation after line validation",
+  "title": "Line update: improve navigation after update to match user history",
   "type": "fix",
   "packages": "stock"
 }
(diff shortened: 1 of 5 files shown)


packages/apps/stock/src/components/templates/customerDelivery/CustomerDeliveryLineButtons/CustomerDeliveryLineButtons.js
```diff
@@ -54,12 +56,19 @@ const CustomerDeliveryLineButtons = ({
       }),
     );
 
-    navigateBackToDetails();
+    if (isScreenMounted('CustomerDeliveryLineListScreen')) {
+      navigation.pop();
```
lme-axelor, line 60: navigation pop won't work in all cases as depending on the configuration, you can be coming from the line verification screens
You need to navigate back to the lines list screen directly

changelogs/unreleased/92404.json
```diff
@@ -0,0 +1,5 @@
+{
+  "title": "Navigation stock move line: improve navigation after line validation",
```
lme-axelor, line 2: You forgot to add this behavior on the inventory lines

vhu-axelor (approved):

PR #950, opened 2025-03-31
feat: add messenger header action on user screen

vhu-axelor:
RM#92648

### DIFF
--- a/packages/apps/message/src/hooks/use-message-header-actions.tsx
+++ b/packages/apps/message/src/hooks/use-message-header-actions.tsx
@@ -37,6 +37,7 @@ import {useIsSubscribed} from './use-is-subscribed';
 export const useMessageHeaders = () => {
   useMailMessagesGenericAction();
   useMailMessagesDetailsAction();
+  useUserProfileActions();
 };
 
 const useMailMessagesGenericAction = () => {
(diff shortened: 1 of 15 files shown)


packages/apps/message/src/hooks/use-message-header-actions.tsx
```diff
@@ -109,3 +110,23 @@ const useMailMessagesDetailsAction = () => {
...
+  const I18n = useTranslator();
+  const navigation = useNavigation();
+
+  useEffect(() => {
+    headerActionsProvider.registerModel('auth_user_profile', {
+      actions: [
+        {
+          key: 'inbox',
+          order: 15,
+          iconName: 'bell-fill',
```
lme-axelor, line 124: Use [chat-dots](https://icons.getbootstrap.com/icons/chat-dots/) like on the web instead.

packages/apps/message/src/screens/InboxScreen.tsx
```diff
@@ -0,0 +1,72 @@
...
+    [dispatch],
+  );
+
+  return (
+    <Screen>
+      <ScrollList
+        loadingList={loadingInbox}
+        data={inboxList}
+        renderItem={({item}) => (
+          <MailMessageCard
```
lme-axelor, line 45: There are some missing information on the messages like which record is updated/created like it's done on the web. We need to have the same informations

PR #953, opened 2025-04-04
feat: add WebSocket support

vhu-axelor:
RM#93320

PR #957, opened 2025-04-07
feat: improve header action with unread messages count

vhu-axelor:
RM#92649


vhu-axelor:
Waiting for PRs #950 and #953

PR #960, opened 2025-04-08
feat: manage attached files through dms package

vhu-axelor:
RM#92656


packages/ui/src/components/molecules/index.tsx
```diff
@@ -17,7 +17,6 @@
  */
 
 export {default as Alert} from './Alert/Alert';
-export {default as AttachmentCard} from './AttachmentCard/AttachmentCard';
```
lme-axelor, line 20: Add a changelog to inform that we deleted this component

packages/ui/src/components/molecules/AttachmentCard/AttachmentCard.tsx
```diff
@@ -1,85 +0,0 @@
...
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- * GNU Affero General Public License for more details.
- *
- * You should have received a copy of the GNU Affero General Public License
- * along with this program.  If not, see <http://www.gnu.org/licenses/>.
- */
-
-import React from 'react';
-import {StyleSheet, TouchableOpacity, View} from 'react-native';
-import File from '../../../types/file';
```
lme-axelor, line 21: I think this type can be deleted too

packages/core/src/screens/index.js
```diff
@@ -18,17 +18,9 @@
 
 import BarcodeDisplayScreen from './BarcodeDisplayScreen';
 import JsonFieldScreen from './JsonFieldScreen';
-import MailMessageAttachedFilesScreen from './MailMessageAttachedFilesScreen';
 import ProcessListScreen from './ProcessListScreen';
 
 export default {
-  MailMessageAttachedFilesScreen: {
-    title: 'Base_AttachedFiles',
```
lme-axelor, line 26: This translation can be deleted

packages/core/src/screens/MailMessageAttachedFilesScreen.tsx
```diff
@@ -1,135 +0,0 @@
...
-
-const MailMessageAttachedFilesScreen = ({route}) => {
-  const {files} = route.params;
-  const Colors = useThemeColor();
-  const I18n = useTranslator();
-  const dispatch = useDispatch();
-
-  const {baseUrl, token, jsessionId} = useSelector(state => state.auth);
-  const {loading, attachedFilesList} = useSelector(
-    state => state.attachedFiles,
```
lme-axelor, line 42: This slice too ?

packages/apps/dms/src/api/document-api.js
```diff
@@ -104,6 +105,14 @@ const createDocumentCriteria = ({
     });
   }
 
+  if (Array.isArray(attachedFileIds) && attachedFileIds.length > 0) {
+    criteria.push({
+      fieldName: 'metaFile.id',
+      operator: 'in',
+      value: attachedFileIds,
+    });
+  }
```
lme-axelor, line 114: I think the dms package should only add this possibility to search document through a list of id (without notion of attached files)
Everything else should be managed through the message package but no need to add a new slice searchDocument will be sufficient to fetch data

Add a changelog for that addition

packages/apps/dms/src/screens/index.ts
```diff
@@ -51,9 +52,17 @@ export default {
       shadedHeader: false,
     },
   },
+  MailMessageAttachedFilesScreen: {
+    title: 'Dms_AttachedFiles',
+    component: MailMessageAttachedFilesScreen,
+    options: {
+      shadedHeader: false,
+    },
+  },
```
lme-axelor, line 61: This is not consistent as it's message which depends on dms but dms screen is named after message  objects.
This screen should be build inside the message package and use components/functions/tools from the dms package. Else the dependency inside the message package is not really useful.

packages/ui/src/components/molecules/AttachmentCard/AttachmentCard.tsx
vhu-axelor: No we can't because it's used in ui by file downloader tool

lme-axelor: To get file extension, I think it can be transformed to a simple utils function or something. The rest of the type is not used and the getFileExtension function is quite simple...

vhu-axelor: Yes I can move the getFileExtension logic and delete the type file...

PR #984, opened 2025-04-28
feat: add responses management

vhu-axelor:
RM#92651

### DIFF
--- a/packages/apps/message/src/api/mail-message-api.ts
+++ b/packages/apps/message/src/api/mail-message-api.ts
@@ -246,3 +246,9 @@ export async function fetchInboxMessages({
     url: `/ws/messages?folder=inbox&limit=${limit}&offset=${limit * page}`,
   });
 }
+
+export async function fetchReplies({messageId}: {messageId: number}) {
+  return axiosApiProvider.get({
+    url: `/ws/messages/${messageId}/replies`,
+  });
+}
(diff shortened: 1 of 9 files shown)


packages/apps/message/src/api/mail-message-api.ts
```diff
@@ -246,3 +246,9 @@ export async function fetchInboxMessages({
     url: `/ws/messages?folder=inbox&limit=${limit}&offset=${limit * page}`,
   });
 }
+
+export async function fetchReplies({messageId}: {messageId: number}) {
+  return axiosApiProvider.get({
+    url: `/ws/messages/${messageId}/replies`,
```
lme-axelor, line 252: Can you check if the pagination is managed like it's done on the classic messages ?

packages/apps/message/src/components/organisms/MailMessageCard/MailMessageCard.tsx
```diff
@@ -56,58 +60,137 @@ const MailMessageCard = ({
...
+                {numReplies}
+              </Text>
+              <Icon
+                name="chat-dots"
+                touchable
+                onPress={() => {
+                  getReplies();
+                  setAreRepliesVisible(current => !current);
+                }}
+              />
```
lme-axelor, line 101: Make this whole part touchable
And please check which is better between text/icon and icon/text display

lme-axelor (commented):
Please try to improve margin management on the reply. Currently too much on the left and not enough on the right

packages/apps/message/src/api/mail-message-api.ts
vhu-axelor: I tested it and I don't think so

PR #989, opened 2025-05-07
Wip feat 92653

vhu-axelor:
RM#92653


packages/apps/message/src/components/organisms/MailMessageCard/MailMessageCard.tsx
lme-axelor, line 1: Please close input box after sending the message
Instead of displaying the button right under the number of messages, can we display it right under the read message action on the card ? Please check if it can be better or else we need to find another solution, currently it's too much information on the side and not very understandable.

If the replies are not opened, after sending the message it should just display the new message and not all replies

PR #996, opened 2025-06-02
feat: add QI form view

gca-axelor:
RM#93804


vhu-axelor:
I forgot to write it in my review but if you select system type with analysis detection, step 2 is empty

PR #1003, opened 2025-06-13
feat: add management of important messages

vhu-axelor:
RM#95557


gca-axelor:
nice works !, just an issue when you open inbox in a small message , the box goes below the message after

gca-axelor:
nice just need to manage sub message

PR #1004, opened 2025-06-13
feat: add management of archived messages

vhu-axelor:
RM#95558


vhu-axelor:
Waiting for PR#1003

gca-axelor:
nice works ! just need to mangage to unarchived message

PR #1025, opened 2025-07-04
feat: add management of Zebra keyboard entries

vhu-axelor:
RM#96873


vhu-axelor:
⚠️ POC
This is just a quick POC to capture the key codes from the physical keyboard of the Zebra device (and any other keyboard). The code isn’t clean or optimized yet — it will need to be refactored and turned into a proper library in the future.

PR #1029, opened 2025-07-09
refactor: use rntl test for ScrollView component

gca-axelor:
RM#97821


gca-axelor:
i can't use toHaveStyle for customStyle because the property use for style is "contentContainerStyle"

PR #1073, opened 2025-08-27
feat: add payment management on saleOrder creation

gca-axelor:
rm#97962


packages/apps/sale/src/components/organisms/PaymentModeSearchBar/PaymentModeSearchBar.tsx
```diff
@@ -0,0 +1,87 @@
...
+const PaymentModeSearchBar = ({
+  style = null,
+  title = 'Sale_PaymentMode',
+  showTitle = false,
+  defaultValue = null,
+  onChange = () => {},
+  readonly = false,
+  required = false,
+  companyId: _companyId,
+}: PaymentModeSearchBarProps) => {
```
lme-axelor, line 49: Same comments

packages/apps/sale/src/features/paymentConditionSlice.js
```diff
@@ -0,0 +1,59 @@
...
+  initialState,
+  extraReducers: builder => {
+    generateInifiniteScrollCases(builder, searchPaymentCondition, {
+      loading: 'loading',
+      moreLoading: 'moreLoading',
+      isListEnd: 'isListEnd',
+      list: 'paymentConditionList',
+    });
+  },
+});
```
lme-axelor, line 57: We need to manage the case where the module account does not exist in the ERP.

To test : add a few letter inside the model name so you get a ClassNotFoundException. The objective is not make it invisible for the user and to not display both fields if the module does not exist

Solution 1 : In case of this error, the list should be set to empty and the searchbar should be hidden ( last part should be managed inside the specific search bar component)

Solution 2 : Create a tool inside the core package to check if a web package is installed.
Maybe by asking for an App (com.axelor.studio.db.App) name and checking if it's installed
Check installation before doing the API call and do not display the payment mode & condition if the module is not installed

Solution 2 is cleaner and could be reuse later so try that first and if it's too complicated, try solution 1

